PR #7, opened 2015-02-05
Fix :Rust* commands

kballard:
Fix a number of issues with the `:RustRun`, `:RustExpand`,
`:RustEmitAsm`, and `:RustEmitIr` commands.

### DIFF
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -175,7 +175,7 @@ function! s:Emit(dict, type, args)
 			silent put =output
 			1
 			d
-			if a:type == "ir"
+			if a:type == "llvm-ir"
 				setl filetype=llvm
 				let extension = 'll'
 			elseif a:type == "asm"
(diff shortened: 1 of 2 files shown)


autoload/rust.vim
```diff
@@ -208,6 +280,23 @@ function! rust#AppendCmdLine(text)
 	return cmd
 endfunction
 
+" Tokenize the String according to shell parsing rules
+function! s:ShellTokenize(text)
```
chris-morgan: I’m not confident about this function across different shells. The most likely culprit for problems is Windows, which has a very basic shell as far as things like interpolation and quotation marks are concerned—has it been tested there?

kballard: Good point, I only considered sh-compatible shells (as that implements the sh tokenizing rules). Perhaps the best bet would be to also remove all the escape characters during tokenizing, so it can be passed back to `shellescape()`. It could theoretically implement platform-native tokenizing rules but I don't have a clue what the Windows shell rules are, and personally, I don't see a problem with just defining it as the sh rules.

kballard:
@chris-morgan I've implemented a new shell tokenizer. In my quick tests it seems to work, but it could use another pair of eyes.

kballard:
@chris-morgan r?

PR #9, opened 2015-02-13
Replace `be` with `become`

ranma42:
The reserved keyword `be` has been replaced with `become` by
rust-lang/rust#21918.


steveklabnik:
Thank you!

PR #14, opened 2015-03-04
Include ctags definitions to support Tagbar

ches:
If the popular [Tagbar](https://github.com/majutsushi/tagbar) plugin is in use, set it up with the necessary settings to support Rust.

### DIFF
--- a/ftplugin/rust/tagbar.vim
+++ b/ftplugin/rust/tagbar.vim
@@ -1,0 +1,32 @@
+"
+" Support for Tagbar -- https://github.com/majutsushi/tagbar
+"
+if !exists(':Tagbar')
+  finish
+endif
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+let g:tagbar_type_rust = {
+  \ 'ctagstype' : 'rust',
+  \ 'kinds' : [
+    \'T:types',
+    \'f:functions',
+    \'g:enumerations',
+    \'s:structures',
+    \'m:modules',
+    \'c:constants',
+    \'t:traits',
+    \'i:trait implementations',
+  \ ]
+\ }
+
+" In case you've updated/customized your ~/.ctags and prefer to use it.
+if get(g:, 'rust_use_builtin_ctags_defs', 1)
+  let g:tagbar_type_rust.deffile = expand('<sfile>:p:h:h:h') . '/ctags/rust.ctags'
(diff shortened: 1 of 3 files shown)


ftplugin/rust/tagbar.vim
```diff
@@ -0,0 +1,32 @@
...
+  \ 'ctagstype' : 'rust',
+  \ 'kinds' : [
+    \'T:types',
+    \'f:functions',
+    \'g:enumerations',
+    \'s:structures',
+    \'m:modules',
+    \'c:constants',
+    \'t:traits',
+    \'i:trait implementations',
```
dotdash: `d:macros` seems to be missing here

dotdash:
Thanks! Seems that I will be able to drop those bits from my vimrc :-) The PR LGTM (with the macro type added), but my VimL-fu is too weak to r+

ches:
Ah yes, macros should be part of the Tagbar kinds list, I'll fix that. Thanks!

brson:
@ches Sorry this fell off the radar. Is it still relevant and ready to go?

ches:
It's my fault for not following up with the addition of macros—let me add that and make sure it's all good, once I have a few minutes for it.

tbelaire:
This is missing `const` and only has `static` for global variables.  You can apply the [same fix][1] as I gave to rusty-tags, change the static line to:

    --regex-Rust=/^[ \\t]*(pub[ \\t]+)?(static|const)[ \\t]+([a-zA-Z0-9_]+)/\\3/c,consts,static constants/"

[1]: https://github.com/dan-t/rusty-tags/pull/15/files

steveklabnik:
Ping! What's the current state of this PR?

da-x:
Hi @ches and all, 

It has been awhile, and I am under the impression that https://github.com/dan-t/rusty-tags is maintained and in use by many.

However I do deem this PR a worthy contribution and want to integrate it, with a few changes:

1) A default of `g:rust_use_builtin_ctags_defs = 0`.
2) A small section on the main README.md that lets the user decide on whether to use the quick builtin or install `rusty-tags`.

Sounds good?

da-x:
Merged, and verified that rusty-tags can be used in conjunction with Tagbar. I also changed the default for Tagbar activation.

Thanks!

PR #16, opened 2015-03-25
Add rudimentary support for the playpen

richo:
This command adds :Play, which will post your current buffer to the
playpen and give you a sharable link.

### DIFF
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -363,42 +363,54 @@ endfunction
 " Parts of gist.vim by Yasuhiro Matsumoto <mattn.jp@gmail.com> reused
 " gist.vim available under the BSD license, available at
 " http://github.com/mattn/gist-vim
-if globpath(&rtp, 'autoload/webapi/http.vim') !=# ''
-  " Only setup playpen support if we have webapi
-  if !exists('g:rust_playpen_url')
-    let g:rust_playpen_url = 'https://play.rust-lang.org/'
-  endif
-  if !exists('g:rust_shortener_url')
-    let g:rust_shortener_url = 'https://is.gd/'
-  endif
-
-  function! rust#Play(count, line1, line2, ...) abort
-    redraw
-    let bufname = bufname('%')
-    if a:count < 1
-      let content = join(getline(a:line1, a:line2), "\n")
-    else
-      let save_regcont = @"
-      let save_regtype = getregtype('"')
-      silent! normal! gvy
-      let content = @"
-      call setreg('"', save_regcont, save_regtype)
-    endif
-
-    let body = g:rust_playpen_url."?code=".webapi#http#encodeURI(content)
-
-    if strlen(body) > 5000
-      echohl ErrorMsg | echomsg 'Buffer too large, max 5000 encoded characters ('.strlen(body).')' | echohl None
-      return
-    endif
-
-    let payload = "format=simple&url=".webapi#http#encodeURI(body)
-    let res = webapi#http#post(g:rust_shortener_url.'create.php', payload, {})
-    let url = res.content
-    redraw | echomsg 'Done: '.url
-  endfunction
-
+if !exists('g:rust_playpen_url')
+  let g:rust_playpen_url = 'https://play.rust-lang.org/'
 endif
+if !exists('g:rust_shortener_url')
+  let g:rust_shortener_url = 'https://is.gd/'
+endif
+
+function! s:has_webapi()
+  if !exists("*webapi#http#post")
+    try
+      call webapi#http#post()
+    catch
+    endtry
+  endif
+  return exists("*webapi#http#post")
+endfunction
+
+function! rust#Play(count, line1, line2, ...) abort
+  redraw
+
+  if !s:has_webapi()
+    echohl ErrorMsg | echomsg ':RustPlay depends on webapi.vim (https://github.com/mattn/webapi-vim)' | echohl None
+    return
+  endif
+
+  let bufname = bufname('%')
+  if a:count < 1
+    let content = join(getline(a:line1, a:line2), "\n")
+  else
+    let save_regcont = @"
+    let save_regtype = getregtype('"')
+    silent! normal! gvy
+    let content = @"
+    call setreg('"', save_regcont, save_regtype)
+  endif
+
+  let body = g:rust_playpen_url."?code=".webapi#http#encodeURI(content)
+
+  if strlen(body) > 5000
+    echohl ErrorMsg | echomsg 'Buffer too large, max 5000 encoded characters ('.strlen(body).')' | echohl None
+    return
+  endif
+
+  let payload = "format=simple&url=".webapi#http#encodeURI(body)
+  let res = webapi#http#post(g:rust_shortener_url.'create.php', payload, {})
+  let url = res.content
+  redraw | echomsg 'Done: '.url
+endfunction
 
 " }}}1
 
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -405,6 +405,7 @@ function! rust#Play(count, line1, line2, ...) abort
   let payload = "format=simple&url=".webapi#http#encodeURI(body)
   let res = webapi#http#post(l:rust_shortener_url.'create.php', payload, {})
   let url = res.content
+
   redraw | echomsg 'Done: '.url
 endfunction
 
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -363,13 +363,6 @@ endfunction
 " Parts of gist.vim by Yasuhiro Matsumoto <mattn.jp@gmail.com> reused
 " gist.vim available under the BSD license, available at
 " http://github.com/mattn/gist-vim
-if !exists('g:rust_playpen_url')
-  let g:rust_playpen_url = 'https://play.rust-lang.org/'
-endif
-if !exists('g:rust_shortener_url')
-  let g:rust_shortener_url = 'https://is.gd/'
-endif
-
 function! s:has_webapi()
   if !exists("*webapi#http#post")
     try
@@ -382,6 +375,17 @@ endfunction
 
 function! rust#Play(count, line1, line2, ...) abort
   redraw
+
+  if !exists('g:rust_playpen_url')
+    let l:rust_playpen_url = 'https://play.rust-lang.org/'
+  else
+    let l:rust_playpen_url = g:rust_playpen_url
+  endif
+  if !exists('g:rust_shortener_url')
+    let l:rust_shortener_url = 'https://is.gd/'
+  else
+    let l:rust_shortener_url = g:rust_shortener_url
+  endif
 
   if !s:has_webapi()
     echohl ErrorMsg | echomsg ':RustPlay depends on webapi.vim (https://github.com/mattn/webapi-vim)' | echohl None
@@ -399,7 +403,7 @@ function! rust#Play(count, line1, line2, ...) abort
     call setreg('"', save_regcont, save_regtype)
   endif
 
-  let body = g:rust_playpen_url."?code=".webapi#http#encodeURI(content)
+  let body = l:rust_playpen_url."?code=".webapi#http#encodeURI(content)
 
   if strlen(body) > 5000
     echohl ErrorMsg | echomsg 'Buffer too large, max 5000 encoded characters ('.strlen(body).')' | echohl None
@@ -407,7 +411,7 @@ function! rust#Play(count, line1, line2, ...) abort
   endif
 
   let payload = "format=simple&url=".webapi#http#encodeURI(body)
-  let res = webapi#http#post(g:rust_shortener_url.'create.php', payload, {})
+  let res = webapi#http#post(l:rust_shortener_url.'create.php', payload, {})
   let url = res.content
   redraw | echomsg 'Done: '.url
 endfunction
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -376,16 +376,8 @@ endfunction
 function! rust#Play(count, line1, line2, ...) abort
   redraw
 
-  if !exists('g:rust_playpen_url')
-    let l:rust_playpen_url = 'https://play.rust-lang.org/'
-  else
-    let l:rust_playpen_url = g:rust_playpen_url
-  endif
-  if !exists('g:rust_shortener_url')
-    let l:rust_shortener_url = 'https://is.gd/'
-  else
-    let l:rust_shortener_url = g:rust_shortener_url
-  endif
+  let l:rust_playpen_url = get(g:, 'rust_playpen_url', 'https://play.rust-lang.org/')
+  let l:rust_shortener_url = get(g:, 'rust_shortener_url', 'https://is.gd/')
 
   if !s:has_webapi()
     echohl ErrorMsg | echomsg ':RustPlay depends on webapi.vim (https://github.com/mattn/webapi-vim)' | echohl None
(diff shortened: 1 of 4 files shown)


autoload/rust.vim
```diff
@@ -221,5 +221,47 @@ function! s:RmDir(path)
 endfunction
 
 " }}}1
```
chris-morgan: This `" }}}1` should be removed—it’s only desirable after the last one.

richo: Fixed

richo:
Updated to add some documentation, and hints about how to make Play go if you're not using webapi.vim

r? @chris-morgan

doc/rust.txt
```diff
@@ -157,6 +157,14 @@ COMMANDS                                                       *rust-commands*
 		If |g:rustc_path| is defined, it is used as the path to rustc.
 		Otherwise it is assumed rustc can be found in $PATH.
 
+:Play                                                                   *Play*
```
chris-morgan: `*Play*` should be `*:Play*`.

richo: Oops! Fixed.

doc/rust.txt
```diff
@@ -157,6 +157,14 @@ COMMANDS                                                       *rust-commands*
...
 		Otherwise it is assumed rustc can be found in $PATH.
 
+:Play                                                                  *:Play*
+        This command is only available if you have web-api.vim[0]
+        installed.  It sends the current selection, or if nothing is
+        selected, the entirety of the current buffer to the Rust playpen,
+        and emits a message with the is.gd shortened URL to the playpen.
+
+        [0]: https://github.com/mattn/webapi-vim
+
```
chris-morgan: This file is (as is standard for Vim help) formatted with eight-character-wide tabs; your four spaces will need to be replaced and the paragraph reflowed to the textwidth of 78. The Markdown-like formatting of the link doesn’t really feel quite right either; how about something like this for the whole thing (looks like GitHub replaces my literal tabs with four spaces; not good; ah well, that part will need fixing):

```
:Play                                                                  *:Play*
		This command is only available if you have web-api.vim
		installed (available at https://github.com/mattn/webapi-vim).
		It sends the current selection, or if nothing is selected, the
		entirety of the current buffer to the Rust playpen, and emits
		a message with the is.gd shortened URL to the playpen.

		|g:rust_playpen_url| is the base URL to the playpen, by
		default "https://play.rust-lang.org/".

```

I also added mention of `g:rust_playpen_url`; it feels like it should be mentioned.

richo:
Ok, I think I finally have this right. I also documented `g:rust_shortener_url`.

doc/rust.txt
```diff
@@ -157,6 +157,20 @@ COMMANDS                                                       *rust-commands*
...
+		This command is only available if you have web-api.vim
+		installed (available at https://github.com/mattn/webapi-vim).
+		It sends the current selection, or if nothing is selected, the
+		entirety of the current buffer to the Rust playpen, and emits a
+		message with the shortened URL to the playpen.
+
+		|g:rust_playpen_url| is the base URL to the playpen, by default
+		"https://play.rust-lang.org/".
+
+		|g:rust_shortener_url| is the base url for the shorterned, by
```
chris-morgan: I wonder what a shorterned is?

richo:
I am not good at compuers today.

autoload/rust.vim
```diff
@@ -220,6 +220,47 @@ function! s:RmDir(path)
...
+      let save_regtype = getregtype('"')
+      silent! normal! gvy
+      let content = @"
+      call setreg('"', save_regcont, save_regtype)
+    endif
+
+    let body = g:rust_playpen_url."?code=".webapi#http#encodeURI(content)
+
+    if strlen(body) > 5000
+      echohl ErrorMsg | echomsg 'Buffer too large, max 5000 encoded characters ('.strlen(body).'(' | echohl None
```
chris-morgan: `s/'('/')'/`

richo:
Just gave it another once over and it looks legit to me now, although I've said that before.

kballard:
I would suggest calling this `:RustPlay` instead.

autoload/rust.vim
```diff
@@ -220,6 +220,47 @@ function! s:RmDir(path)
 	silent exe "!rm -rf " . shellescape(a:path)
 endfunction
 
+" Playpen Support {{{1
+" Parts of gist.vim by Yasuhiro Matsumoto <mattn.jp@gmail.com> reused
+" gist.vim available under the BSD license, available at
+" http://github.com/mattn/gist-vim
+if globpath(&rtp, 'autoload/webapi/http.vim') !=# ''
```
kballard: This is not compatible with lazy bundle loading with [NeoBundle](https://github.com/Shougo/neobundle.vim). An alternative approach that would work might look like

```vim
function! s:has_webapi()
    if !exists("*webapi#http#post")
        try
            function("webapi#http#post")
        catch
        endtry
    endif
    return exists("*webapi#http#post")
endfunction
```

The downside to this is it triggers a load of `webapi`, and you don't actually need it to be loaded until the user runs `:Play`. So I'd suggest doing this test at the top of `rust#Play`. Doing that also means you can unconditionally define `:Play` and have it emit an error telling the user they need to install `webapi` if it's not installed.

richo:
I'm not averse to renaming it. Rationale was basically that this should be analogous to `:Gist` that I use all the time, but I'll rename it in a second.

autoload/rust.vim
richo: Sure, this is much cleaner. Thanks!

richo:
@kballard I changed it to work out if it works at runtime. I had to change your example a bit, `function()` didn't cause the code to get loaded for me (I use pathogen), but just calling it without args did.

Will this work for NeoBundle users?

autoload/rust.vim
```diff
@@ -220,6 +220,59 @@ function! s:RmDir(path)
...
+" Playpen Support {{{1
+" Parts of gist.vim by Yasuhiro Matsumoto <mattn.jp@gmail.com> reused
+" gist.vim available under the BSD license, available at
+" http://github.com/mattn/gist-vim
+if !exists('g:rust_playpen_url')
+  let g:rust_playpen_url = 'https://play.rust-lang.org/'
+endif
+if !exists('g:rust_shortener_url')
+  let g:rust_shortener_url = 'https://is.gd/'
+endif
```
richo: Err, I guess these should be local to the script now. Sec.

chris-morgan:
Current functionality LGTM with pathogen.

For further work, though: echoing is useful for some situations but not all—the user may not be able to select it or open it from that location, for example. I think that the ability to dump it to a stipulated register would be handy; something along these lines (variable names in particular up for bikeshedding):

```vim
if exists('g:rust_play_url_register')
    let url_register = g:rust_play_url_register
else
    let url_register = '"'
endif
```
```vim
if url_register != ""
    exec "let @" . url_register "= url"
endif
```

So by default it would put the URL in the default register, and it can be turned off by assigning an empty value to the global, and the documentation would suggest "+" as a particularly useful value.

richo:
So this code was pretty directly ported out of `Gist.vim`- which regularly drops the gist url from the view, so I just use `:mess` to bring up the message list to copy it out.

There are plenty of other things this could do, if nothing else there's just a lot of code that I could have copied over, but figured I'd start minimal and maybe add features in future if they're actually necessary/wanted.

richo:
Are there final thoughts on how much this should bring in initially? while there's definitely a part of me that considered just making Gist.vim a bit more generic so I could consume it's primitives here, I kinda left it to YAGNI, decided to implement the bare minimum amount needed to be useful and figured bolting on more stuff later is less painful than tiptoeing around code that's not actually used by anyone.

richo:
ping. @steveklabnik you're the r? on this, so hassling you.

steveklabnik:
I am terrible at vimscript. This is a good feature to have, but I'm not a good reviewer in terms of 'is this quality'

richo:
So my belief is that the current code has been looked at, and the remainder is bikeshedding about how much stuff to implement in the first cut.

tsion:
I was just wishing for this functionality. Anything we can do to speed up the merge? It looks like this won't affect anyone not trying to use the command, even if something isn't quite right, so it should be pretty safe to merge.

I'd rather see it go in as-is as soon as possible than waiting for any more features.

richo:
It seems broken in my environment, even though when I use a local handler for the request and send the *exact same bytes* to is.gd, I get a valid response back. I would love for someone else to test this and confirm that I've just done something tremendous to my environment.

@tsion are you interested in testing? Have you already?

tsion:
I just switched to using your branch locally and it's working as intended for me.

However, I noticed it doesn't put the URL in my system clipboard like gist.vim does - is this something you're interested in adding? (Now I'm bikeshedding, too.)

richo:
I have no issue adding it. The underlying problem is that generic vimscript is hard, and there's no obvious way for me to just frob a few bits in gist.vim and making it do something mostly the same but very slightly different, which is basically why I copied as little code as possible.

If this lands, i'll work out how to deal with the clipboard. Does that seem ok?

tsion:
Yeah, I'd still rather this got merged pre-clipboard support. Thanks for your work on this. :heart:

kballard:
> I changed it to work out if it works at runtime. I had to change your example a bit, function() didn't cause the code to get loaded for me (I use pathogen), but just calling it without args did.

> Will this work for NeoBundle users?

Oh you're right, I misinterpreted the results of the `function()` function. It always hands back a funcref for any `foo#bar` name even if the name is invalid, without triggering autoload. I think calling `webapi#http#post()` in a `try` block like that should be sufficient; the function is declared as taking a required argument `url`, so that call is guaranteed to fail without doing anything (beyond autoloading the file).

ftplugin/rust.vim
```diff
@@ -106,6 +106,9 @@ command! -nargs=* -bar -buffer RustEmitIr call rust#Emit("ir", [<f-args>])
 " See |:RustEmitAsm| for docs
 command! -nargs=* -bar -buffer RustEmitAsm call rust#Emit("asm", [<f-args>])
 
+" See |:RustPlay| for docs
+command! -range=% RustPlay :call rust#Play(<count>, <line1>, <line2>, <f-args>)
```
kballard: You need to add the appropriate `delcommand` to `b:undo_ftplugin`. See [line 144](#diff-9f6aaa9ba985bf46731062ecf337d32aR144).

kballard:
Besides the `delcommand` I think this is ok to merge. It can't be automatically merged right now but the conflict is trivial to resolve.

As for the clipboard, I'd like to see that added too, with a config variable to turn it off (default to on). mattn/gist-vim appears to have some logic for invoking external commands to modify the pasteboard, but I think that's unnecessary here; as long as vim has the `clipboard` feature you can just write to the `+` register. I think it's sufficient to just document that this aspect requires `+clipboard`.

richo:
Updated the patch. I'll fiddle a bit with the clipboard at some point, but no promises on when.

autoload/rust.vim
```diff
@@ -359,6 +359,63 @@ function! s:system(pwd, cmd)
...
+  if !exists('g:rust_playpen_url')
+    let l:rust_playpen_url = 'https://play.rust-lang.org/'
+  else
+    let l:rust_playpen_url = g:rust_playpen_url
+  endif
+  if !exists('g:rust_shortener_url')
+    let l:rust_shortener_url = 'https://is.gd/'
+  else
+    let l:rust_shortener_url = g:rust_shortener_url
+  endif
```
kballard: FWIW, this can be expressed more simply as

```vim
let l:rust_playpen_url = get(g:, 'rust_playpen_url', 'https://play.rust-lang.org/')
let l:rust_shortener_url = get(g:, 'rust_shortener_url', 'https://is.gd/')
```

doc/rust.txt
```diff
@@ -157,6 +157,20 @@ COMMANDS                                                       *rust-commands*
...
+		(available at https://github.com/mattn/webapi-vim).  It sends the
+		current selection, or if nothing is selected, the entirety of the
+		current buffer to the Rust playpen, and emits a message with the
+		shortened URL to the playpen.
+
+		|g:rust_playpen_url| is the base URL to the playpen, by default
+		"https://play.rust-lang.org/".
+
+		|g:rust_shortener_url| is the base url for the shorterner, by
+		default "https://is.gd/"
```
kballard: These two settings should be independently documented in the SETTINGS section.

kballard:
At this point I think it's ok to merge, as long as the clipboard thing and the documentation tweak will be submitted as a separate PR (although I won't object if the documentation tweak is added to this one).

ping @steveklabnik

kballard:
When this does get merged, any outstanding work should be captured in an open issue.

richo:
I fixed the flag parsing, and added docs in SETTINGS.

I can open an issue for the clipboard things.

doc/rust.txt
```diff
@@ -95,6 +95,19 @@ g:cargo_manifest_name~
 	    let g:cargo_manifest_name = 'Cargo.toml'
 <
 
+                                                          *g:rust_playpen_url*
+g:rust_playpen_url~
+	Set this option to override the url for the playpen to use:
+	    let g:cargo_manifest_name = 'Cargo.toml' >
```
ches: Copypasta oversight—this example is not for the config var being documented here :grin:

richo: oof, fixed.

kballard:
:+1:

richo:
Ping. Can this merge?

autoload/rust.vim
```diff
@@ -359,6 +359,56 @@ function! s:system(pwd, cmd)
...
+  let body = l:rust_playpen_url."?code=".webapi#http#encodeURI(content)
+
+  if strlen(body) > 5000
+    echohl ErrorMsg | echomsg 'Buffer too large, max 5000 encoded characters ('.strlen(body).')' | echohl None
+    return
+  endif
+
+  let payload = "format=simple&url=".webapi#http#encodeURI(body)
+  let res = webapi#http#post(l:rust_shortener_url.'create.php', payload, {})
+  let url = res.content
```
ches: It seems so unlikely that another URL shortener is going to work in exactly the same way as is.gd (same params, the path of `create.php`) that I wonder if making and documenting `rust_shortener_url` as a variable is truly useful.

Perhaps it would it would make sense to allow users to provide a custom function with a certain name which should return the final URL, defaulting to the is.gd implementation if it a custom one doesn't exist.

richo: I mean, maybe? I'd kinda rather just ship this than bikeshed about it for another few months.

ches: I'd like to have the functionality sooner rather than later too, just giving feedback that came to my mind at the time I happened to review the PR. No need for a snide tone.

richo: Apologies if I came off bluntly. I agree that it's fairly unlikely to have a drop-in thing, I'll see if there's a convenient way to make it play nicely with other shorteners.

kballard: I agree it would be nice to make this configurable enough that other shorteners could be used instead of just drop-in replacements. But I don't think that needs to hold up this PR (although it might make sense to un-document `g:rust_shortener_url` with the expectation that such a change would use a different configuration).

tsion:
So is it agreed that nothing is now holding up this PR? I'm still using an old branched version of rust.vim that has playpen support, but I'd like to see this merged and move up to master.

brson:
This seems to have fallen off the radar. Based on the general tone of enthusiasm I took the leap to merge it. Let me know if I screwed up.

PR #28, opened 2015-05-18
Skip -> in matchparen for better Fn generics

### LINKED ISSUE #25: Generic parameter lists containing Fn traits not highlighted correctly
Currently, if you use a type like `Box<Fn(A) -> B>` the latter part of the arrow is misinterpreted as the closing bracket of the generic parameter list. This leads to an incorrect matching, so that the wrong bracket is highlighted. Can this be fixed by this plugin?

chris-morgan:
If anyone was overriding the highlighting for `rustArrow`, they will need to change it to `rustArrowCharacter`.

Special thanks to FDinoff for [identifying the solution](http://stackoverflow.com/a/30309949/497043) and @aepsil0n for *asking* for a solution!

Fixes #25.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -116,7 +116,9 @@ syn match     rustSigil        display /[&~@*][^)= \t\r\n]/he=e-1,me=e-1
 " This isn't actually correct; a closure with no arguments can be `|| { }`.
 " Last, because the & in && isn't a sigil
 syn match     rustOperator     display "&&\|||"
-syn match     rustArrow        display "->"
+" This is rustArrowCharacter rather than rustArrow for the sake of matchparen,
+" so it skips the ->; see http://stackoverflow.com/a/30309949 for details.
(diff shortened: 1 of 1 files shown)


fdinoff:
You also need to change line 92 of `ftplugin/rust.vim`

    let b:match_skip = 's:comment\|string\|rustArrow'

to use `rustArrowCharacter`. This will allows `%` to work properly.

chris-morgan:
@fdinoff: skipping uses `=~`, so the fact that it still has the prefix `rustArrow` is sufficient.

aepsil0n:
btw, while highlighting works correctly for the brackets now, when you hover over the arrow it still highlights the opening bracket. A comparatively minor issue, but still, I wonder why that is.

chris-morgan:
@aepsil0n: I believe that part is unavoidable. You can observe it in other filetypes too, *e.g.* `{ "}" }` in C.

PR #29, opened 2015-05-22
Added snippets for UltiSnips

RomanTsegelskyi:
Create a folder for snippets. Added snippets for [UltiSnips](https://github.com/SirVer/ultisnips)


chris-morgan:
I’ve thought about this at length and decided that the main rust.vim repository isn’t a good place for this; it’s a little subjective and not really related to the core Rust/Vim experience. I suggest you maintain a repository of your own for the purpose—it’ll be focused on that and can then get contributions directly to it if people are interested in it.

If others feel that integrating this here would be a good idea, we can reopen it.

PR #32, opened 2015-05-28
update install with NeoBundle

mozillazg:
When I execute `:NeoBundleInstall`, I got: 

    [neobundle/install] You may have used the wrong bundle name, or all of the bundles are already installed.


steveklabnik:
Thanks!

PR #34, opened 2015-06-01
Add match and else to the keywords

IvanUkhov:
The `match` and `else` keywords are currently missing proper highlighting.

Regards,
Ivan

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -23,7 +23,7 @@ syn keyword   rustKeyword     box nextgroup=rustBoxPlacement skipwhite skipempty
 syn keyword   rustKeyword     continue
 syn keyword   rustKeyword     extern nextgroup=rustExternCrate,rustObsoleteExternMod skipwhite skipempty
 syn keyword   rustKeyword     fn nextgroup=rustFuncName skipwhite skipempty
-syn keyword   rustKeyword     for in if impl let
+syn keyword   rustKeyword     for in match if else impl let
 syn keyword   rustKeyword     loop pub
 syn keyword   rustKeyword     return super
 syn keyword   rustKeyword     unsafe where while


chris-morgan:
`if`, `else` and `match` are `rustConditional` keywords, and `rustConditional` has default highlighting linked to `Conditional`, which is most commonly highlighted as `Keyword`; the issue you describe suggests that your color scheme is not handling `Conditional` appropriately; this is definitely out of scope for rust.vim.

That `if` is also a `rustKeyword` keyword is a curious and minor issue.

IvanUkhov:
@chris-morgan, thanks for the clarification; absolutely makes sense. Yes, I was primarily motivated by the fact that `if` was in `rustKeyword`.

IvanUkhov:
@chris-morgan, I’ve just checked the [reference](http://doc.rust-lang.org/grammar.html#keywords), and `match` and `else` are considered as keywords. From this standpoint, it also makes sense to include them in both lists, `rustKeyword` and `rustConditional`. A particular color scheme can then decide if it want to color those words that belong to either or both in a special way. Not linking `rustConditional` to Vim’s `Keyword` is bashing a lot of color schemes: conditionals are a special case of keywords, and not all color schemes bother to treat it properly. Thank you.

chris-morgan:
@IvanUkhov This is about syntax highlighting, and in Vim, `if`, `else` and `match` should be `Conditional`, and `for`, `while` and `loop` `Repeat`; `Keyword` is for any other keyword. A keyword will only get one piece of highlighting, and the task is to select the most precise one. Linking `rustConditional` to `Conditional` is the correct thing; the conventions for all of this are detailed in the `group-name` help. You will note that by default `Conditional`, `Repeat` and `Keyword` are all linked to `Statement`—the colour scheme may choose to change that, and Rust should respect its changes. If you’re not happy with that, tweak your colour scheme or `hi link rustConditional Keyword` or similar yourself.

I’ve just started reviewing all this now.

IvanUkhov:
@chris-morgan, I apologize if it sounded offensive; I didn’t mean that. I’m grad that `rust.vim` is in good hands) Thanks!

PR #37, opened 2015-06-23
Make boolean settings more reasonable

EinfachToll:
Currently, only the existence of some of the boolean settings the users can set in their .vimrc are checked, not the actual value of the variable. That means that `let g:rust_conceal = 0` would actually switch on concealing. That's very unintuitive. This PR corrects that.


IanConnolly:
👍

Why prefer `!= 0` over `== 1` out of interest, @EinfachToll?

EinfachToll:
With `!= 0`, this: `let g:rust_conceal = 2` would switch concealing on, while with `== 1`, it would be switched off.
No big deal, but I think that programers are used to the fact that 0 means false and everything else means true.

IanConnolly:
Derp, yes. Sorry, that's what happens when you review code at bed time :D

steveklabnik:
Thanks!

Issue #38, opened 2015-06-25
matchit.vim fails to load with latest neovim


landaire:
[`ftplugin/rust.vim`](https://github.com/rust-lang/rust.vim/blob/master/ftplugin/rust.vim#L93) will source `$VIMRUNTIME/macros/matchit.vim`  regardless of whether or not it needs to be loaded, and always assumes it's in the `macros` dir.

This is not true as of https://github.com/neovim/neovim/pull/2723/. That PR made `matchit.vim` always loaded by default and moved `matchit.vim`: `runtime/macros/matchit.vim → runtime/plugin/matchit.vim`. I believe that a solution would be to instead perform the following:

    if !exists('loaded_matchit')
        source $VIMRUNTIME/macros/matchit.vim
    endif

But I'm still relatively new to vim so I'm not really sure. This *appears* to work on my machine though.

nvim details:

```
~
❯ nvim -v
NVIM 0.0.0-alpha+201506222256 (compiled Jun 23 2015 20:42:16)
Commit: 20a57c497b55bb3632f0d4bcd76c18628cfa5f43
Build type: RelWithDebInfo
Compilation: /usr/local/Library/ENV/4.3/clang -Wconversion -O2 -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -I/tmp/neovim20150623-23404-1k8l993/build/config -I/tmp/neovim20150623-23404-1k8l993/src -I/tmp/neovim20150623-23404-1k8l993/.deps/usr/include -I/tmp/neovim20150623-23404-1k8l993/.deps/usr/include -I/tmp/neovim20150623-23404-1k8l993/.deps/usr/include/luajit-2.0 -I/tmp/neovim20150623-23404-1k8l993/.deps/usr/include -I/tmp/neovim20150623-23404-1k8l993/.deps/usr/include -I/tmp/neovim20150623-23404-1k8l993/.deps/usr/include -I/tmp/neovim20150623-23404-1k8l993/.deps/usr/include -I/Library/Frameworks/Mono.framework/Headers -I/usr/include -I/tmp/neovim20150623-23404-1k8l993/build/src/nvim/auto -I/tmp/neovim20150623-23404-1k8l993/build/include
Compiled by lander@uoeno.local

Optional features included (+) or not (-): +acl   +iconv    +jemalloc
[...]

PR #39, opened 2015-06-28
Add check for matchit being loaded

martinisoft:
This prevents matchit from being loaded again if it is already loaded, which is currently the case for NeoVim users as it is now included by default in the current release and the file is not where the rust plugin will expect matchit to be. This was changed [around here](https://github.com/neovim/neovim/commit/98053f0f9fc3405e48ba3a9089a29fd23b3b414c) in Neovim.

It should resolve rust-lang/rust.vim#38


chris-morgan:
I’m not ready for this to be merged, because it’s the wrong place to put it; NeoVim is at fault here. I’ve filed neovim/neovim#2918 for fixing this where it should be fixed.

martinisoft:
Looks like your issue got a fix merged into master. I'll close this PR. :+1:

PR #49, opened 2015-10-03
Added support for rustfmt before save

grncdr:
I really enjoyed this feature from [vim-go](https://github.com/fatih/vim-go)
so I copied it for rust & rustfmt.

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -1,7 +1,6 @@
-" WTF License
-" Author: stephen@stephensugden.com
+" Author: Stephen Sugden <stephen@stephensugden.com>
 "
-" Directly based on the go#fmt#Format plugin from fatih/vim-go
+" Adapted from https://github.com/fatih/vim-go
 
 if !exists("g:rustfmt_autosave")
   let g:rustfmt_autosave = 0
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -23,16 +23,13 @@ let s:got_fmt_error = 0
 
 function! rustfmt#Format()
   let l:curw = winsaveview()
-  let l:tmpname = tempname()
+  let l:tmpname = expand("%:p:h") . "/." . expand("%:p:t") . ".rustfmt"
   call writefile(getline(1, '$'), l:tmpname)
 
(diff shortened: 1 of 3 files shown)


autoload/rustfmt.vim
```diff
@@ -0,0 +1,80 @@
+" WTF License
```
steveklabnik: This isn't a "real" license, so I wonder if we can take it...

steveklabnik:
General :+1: on this idea, for sure.

grncdr:
I'm happy to update the license, was just a bit lazy when putting this together. The vim-go plugin I was working off of is BSD licensed, this repo has no license, and I guess rust itself is licensed under the Apache & MIT licenses? Would appreciate a recommendation as to the correct license to satisfy all of the above.

A more annoying (for me anyways :wink:) problem: the vim-go plugin works by writing the buffer contents to a temporary file, then calling `gofmt` on that. This doesn't work for rust crates with multiple files, because `rustfmt` will fail when it can't find other modules in the crate. I'm working out a solution, but for now I'd slap a big "WIP" label on this.

steveklabnik:
All Rust stuff is MIT/ Apache2, we forgot to add the license locally when we moved it out of the main repo

grncdr:
Given that there isn't a license on anything else, I've simply removed that line.

IanConnolly:
Was doing this myself!

@grncdr Could you add a `:RustFmt` command that calls the function? No reason why the user couldn't define it themselves.

@steveklabnik What's outstanding for @grncdr to do to get this merged? Happy to help.

steveklabnik:
I'm horrible at vimscript, so I'm not sure. Let's say that yeah, we need that `:RustFmt` command and then let's get this merged. I like the idea of the command so that other stuff can build on top of it, and it seems consistent with the rest of the bindings.

IanConnolly:
@grncdr I know the vim-go function adds errors to the quickfix list, but do you think the location list might be a better option? That way you can keep the errors for multiple windows in their respective location lists. Syntastic adds errors to the loclist for this reason.

oblitum:
So, I guess this feature has been merged in #52, but this pull was left open?

michael-younkin:
Yeah, it looks that way to me.

PR #53, opened 2015-11-13
Create LICENSE

edunham:
https://github.com/rust-lang/rust.vim/issues/45

This would add the Vim license, as was suggested in the issue.

Issue #45, opened 2015-08-17
Please add a LICENSE file


edunham:
Probably dual Apache + MIT like the other Rust stuff?

re: https://github.com/rust-lang/rust/issues/25664

IanConnolly:
For what it's worth, most Vim plugins declare themselves to be distributed under the same terms as Vim itself is.

eg. https://github.com/tpope/vim-fugitive/blame/master/README.markdown#L117-L118

jordwalke:
Any update here?

steveklabnik:
So, we'd have to make it under apache + MIT for now, because that was what it was written under.


If we can get all of these https://github.com/rust-lang/rust.vim/graphs/contributors people to sign off on changing to the vim license, then we can change it. If we think it's worth it, we can do that work.

jordwalke:
Is there anything wrong with Apache+MIT in the mean time?

edunham:
@steveklabnik Thanks for clarifying that it was written under the same terms as Rust (do they apply by default to everything in the Rust org, or something?). Opened new PR to add the same licenses. 

I personally think it's more important to get an open source license into the repo than to bikeshed over which one to use.

steveklabnik:
>  (do they apply by default to everything in the Rust org, or something?)

All the editor stuff was literally in-tree, before I split them out into separate repos, so most of the development was done in rust-lang/rust, hence the licesnse

edunham:
Excellent. #55 landed, so this is now resolved.

PR #55, opened 2015-11-15
License like rust

edunham:
Alternate resolution to https://github.com/rust-lang/rust.vim/issues/45 . If this is accepted,  https://github.com/rust-lang/rust.vim/pull/53 should be closed without merge.


steveklabnik:
Thanks!

Issue #27, opened 2015-05-13
:RustRun doesn't build code properly


dmitry-smith:
https://gist.github.com/dmitry-smith/5f72970e84b38ed38db2

:RustRun does not work for me. It always shows the same error:
main.rs:9:25: 9:41 error: unresolved name `rand::thread_rng`
main.rs:9     let secret_number = rand::thread_rng().gen_range(1, 101);
                                  ^~~~~~~~~~~~~~~~
error: aborting due to previous error 

But when I try to do the same from a command line with "Cargo run" everything is ok. What can I do to fix it? Thx.

habamax:
I have the same issue. It could be that RustRun uses rustc to compile and then run executable.
I think RustRun should use Cargo run first and rustc as a fallback.

habamax:
It looks like we can have
autocmd BufRead,BufNewFile Cargo.toml,Cargo.lock,*.rs compiler cargo
in vimrc and be able to :make run to issue Cargo run.

jfelchner:
If cargo is the blessed way by the Rust community for Rust to be compiled in the 99% case, I think it should be the thing that's executed when `RustRun` is called.  Or at the very least check to see if the `cargo` executable exists.  If it does, use that, if not, use `rustc`.

jdhoek:
Is there a known workaround for this issue?

davidmwalton:
I believe I'm seeing the same (or a closely related) issue. Compiling my app with cargo run causes no issue, but the syntax checking within Vim throws: unresolved name `rand::thread_rng`

Seconding jdhoek's question about whether or not there is a workaround.

Can we possibly squelch known bad errors or something like this? Or otherwise configure rust.vim to use cargo the compile checker vs. rustc?

da-x:
RustRun is a meant as an low-level (and old) way to run Rust code. I will be changing the documentation more to stress that out. Will keep this open in the meanwhile.

heyrict:
I'm currently use the 2018 edition and `RustRun` does not read the `[dependency]` section in `Cargo.toml` and will get me a `E0463: cannot find crate` error.

Currently I'm using a simple command as a replacement. Hope there would be a `:CargoRun` command in the future.
```vim
command -nargs=* RustRun up | execute "!cargo run" <q-args>
```

me-self:
> Currently I'm using a simple command as a replacement. Hope there would be a `:CargoRun` command in the future.

Is `:CargoRun` a command yet?

PR #56, opened 2015-11-15
Fix for screen redrawing issue in RustRun etc.

### LINKED ISSUE #22: screen redrawing issue
There is a very annoying issue with VIM not redrawing on certain TERM values with a dark background after returning from :RustRun. According to the VIM experts,

"this only happens with :! and :shell. use system() instead.  This will avoid the curses mucking that results in the required redraw."

IanConnolly:
`:RustRun` etc. would cause buggy screen redrawing in the error case. Tracked down the issue to a helper function.

I'm not particularly happy about the structure of this functionality. I'd prefer we went through cargo and appropriately handled errors in the quickfix list instead of echohl'ing it out (similar to how we're handling errors in `:RustFmt`) . I'll stick that on my TODO list. (see: #27)

Fixes #22.

### DIFF
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -346,7 +346,7 @@ function! s:RmDir(path)
 		echoerr 'Attempted to delete protected path: ' . a:path
 		return 0
 	endif
-	silent exe "!rm -rf " . shellescape(a:path)
+	return system("rm -rf " . shellescape(a:path))
 endfunction
 
 " Executes {cmd} with the cwd set to {pwd}, without changing Vim's cwd.


IanConnolly:
r? @steveklabnik 

(Chris seems to be have been inactive recently, no sense putting him one the hook for a one-liner)

Jayflux:
Whenever :Rustrun errors i basically get nothing, not even the error on the screen
Then i get the redrawing issue, and i basically have to restart Vim

chris-morgan:
@Jayflux If you ever get drawing issues in Vim, <kbd>Ctrl</kbd>+<kbd>L</kbd> triggers a full screen redraw which fixes such things.

PR #62, opened 2015-12-30
Fix check for g:rust_conceal_mod_path in after/syntax/rust.vim

kballard:
The previous check was quite broken. If `g:rust_conceal_mod_path` wasn't
set, it would throw an error. If it was set to `0`, it would incorrectly
skip linking the highlight groups. It only worked properly for the case
where `g:rust_conceal_mod_path` was set to a non-zero value.

### DIFF
--- a/after/syntax/rust.vim
+++ b/after/syntax/rust.vim
@@ -26,6 +26,6 @@ endif
 
 hi link rustNiceOperator Operator
 
-if !exists('g:rust_conceal_mod_path') && g:rust_conceal_mod_path != 0
+if !(exists('g:rust_conceal_mod_path') && g:rust_conceal_mod_path != 0)
     hi! link Conceal Operator
 endif


Kha:
@steveklabnik Good to merge? This bug makes rust.vim highlighting look quite broken with the default config.

chris-morgan:
What is currently written here makes no sense; that I will definitely agree on.

Using the default highlighting, `->` and `=>` will be highlighted `Operator`, `pub` `Keyword` and `::` `Delimiter`. `Operator` and `Keyword` are often very similar or the same in themes (in molokai which I use (with modifications), `Keyword` is bold and `Operator` is not, but they’re the same colour), but `Delimiter` is usually quite different.

Once using conceal, concealed things can only be of one formatting. With @kballard’s proposed change, if `::`-concealing is turned on then all concealed things are left as the default (which is in my experience *very* ugly, as noted in https://github.com/rust-lang/rust.vim/pull/70#issuecomment-183816951), and if it is not then all are made `Operator`. I think I’d probably prefer removing the conditional as in #70.

Frankly, the whole thing is lipstick on a pig. We’re abusing `conceal`. Removing conceal support from rust.vim seems to me the right thing to do. I have no idea why it got implemented in the first place.

Anyway, for now I suppose we might as well accept this. If someone wants to debate whether the conditional should be removed, go ahead! For my part, I’ll just suggest that conceal support should be removed.

kballard:
I don't think we're abusing `conceal` (Vim allows you to set the conceal replacement characters specifically because this sort of thing is supported), and I don't want to see it removed.

PR #64, opened 2016-01-02
Don't accept unicode escapes without curly braces.

Screwtapello:
At least with Rust 1.5, putting a 4-digit Unicode escape into a string literal produces an error message:

    src/cells.rs:81:19: 81:21 error: incorrect unicode escape sequence
    src/cells.rs:81             (1, "e\u0301"),
				      ^~
    src/cells.rs:81:19: 81:21 help: format of unicode escape sequences is `\u{…}`
    src/cells.rs:81             (1, "e\u0301"),

...and a six-digit Unicode escape (`\Uxxxxxx`) just produces:

    src/cells.rs:81:20: 81:21 error: unknown character escape: U

The only supported Unicode escape sequence is `\u` followed by hex digits in curly braces, so that's the only format the Vim syntax highlighting should accept.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -125,7 +125,6 @@ syn match     rustMacro       '#\w\(\w\)*' contains=rustAssert,rustPanic
 
 syn match     rustEscapeError   display contained /\\./
 syn match     rustEscape        display contained /\\\([nrt0\\'"]\|x\x\{2}\)/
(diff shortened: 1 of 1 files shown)


chris-morgan:
It got changed some time before 1.0, and for compatibility the two schemes coexisted for a while. We just never removed the old way here when it was removed from the parser (which was also before 1.0). Thanks!

PR #66, opened 2016-01-20
Use system instead of systemlist for backwards compatibility.

timothy-mcroy:
Systemlist is not available in some versions of vim.

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -27,7 +27,7 @@ function! rustfmt#Format()
 
   let command = g:rustfmt_command . " --write-mode=overwrite "
 
-  let out = systemlist(command . g:rustfmt_options . " " . shellescape(l:tmpname))
+  let out = split(system(command . g:rustfmt_options . " " . shellescape(l:tmpname)), '\n')
 
   if v:shell_error == 0
     " remove undo point caused via BufWritePre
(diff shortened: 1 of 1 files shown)


chris-morgan:
`systemlist()` was added in 7.4.248, and we do want to support 7.4.0 if convenient.

This splits on just `\n`, not anything like `\r\n`. How does it behave on Windows? If it uses `\r\n` on Windows (uncertain if it does), then I guess this version would end up with the errors’ `text` field having a trailing `\r`. No idea if that matters either. Can someone verify if it all seems sane on Windows?

timothy-mcroy:
Do you feel that matching either pattern would be sufficiently sane?  

``` '\n' ``` would instead be ``` '(\n|\r\n)$' ```

File: autoload/rustfmt.vim [hunk 1]
```diff
@@ -27,7 +27,7 @@ function! rustfmt#Format()
 
   let command = g:rustfmt_command . " --write-mode=overwrite "
 
-  let out = systemlist(command . g:rustfmt_options . " " . shellescape(l:tmpname))
+  let out = split(system(command . g:rustfmt_options . " " . shellescape(l:tmpname)), '(\n|\rn)$')
```
dguo: I think you're missing a backslash for the second n. `\r\n` rather than `\rn`.

PR #70, opened 2016-02-08
Highlights conceals like operators and preserves after colorscheme switch

pyjarrett:
Fixes bug from `if !exists('g:rust_conceal_mod_path') && g:rust_conceal_mod_path != 0` always evaluating to false.

### DIFF
--- a/after/syntax/rust.vim
+++ b/after/syntax/rust.vim
@@ -28,5 +28,5 @@ hi link rustNiceOperator Operator
 hi! link Conceal Operator
 
 " Keeps conceal highlighting proper after return after colorscheme switch.
-autocmd VimEnter,ColorScheme * hi! link Conceal Operator
+autocmd ColorScheme <buffer> hi! link Conceal Operator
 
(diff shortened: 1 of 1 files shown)


chris-morgan:
`VimEnter`? I get `ColorScheme`, but `VimEnter` doesn’t make sense to me. Am I missing something?

An autocmd like this will need to be `<buffer>`.

pyjarrett:
@chris-morgan, I was trying to get it to perform the color command setting initially and then during color scheme swap.  Making it specific to the buffer definitely makes significantly more sense.  Is just doing it on color scheme change sufficient?

I apologize for being green, I've been using vim for a while, but trying to use this to develop more knowledge of VIM customization and plugin development.

pyjarrett:
@chris-morgan, I did some more research on autocmd settings and it looks like ColorScheme is the only condition where this should run.  I made it buffer local as well.

kballard:
I believe the boolean condition that you're deleting was in fact put there for a reason. However, the condition is broken, which my old PR #62 fixes (I have no idea why that still hasn't been merged).

pyjarrett:
@kballard, some (all of the various ones I use) color schemes highlight "concealed" with a bright gray background, while defeats the purpose of concealing. I apologize for missing your pull request, I agree the old boolean condition was broken as it always evaluated to false.  I was under the impression making concealed operators look like operators was the real intent.

PR #72, opened 2016-02-09
Syntax check rust with -Zno-trans instead of -Zparse-only

### LINKED ISSUE #67: Support for file checking beyond syntactic correctness
Currently, the syntastic integration (which will hopefully soon also be used by [neomake](https://github.com/benekastah/neomake/issues/275)) runs `rustc` with `-Zparse-only` to check a particular file. While this is fine, it doesn't catch the vast majority of errors that come up during development.

Obviously, just dropping `-Zparse-only` won't work, as the compiler will complain about missing imports/types, but maybe some kind of integration with `cargo` could work? Maybe look up the file tree to find `Cargo.toml`, run `cargo build`, and parse out any warning/error that pertains to the currently open file? It's not pretty, but would be extremely useful!

faern:
Only parsing missed a lot of errors. This PR changes so more steps of the compile is executed and more errors are found. The solution required some extra code to detect if we are in a standalone rs file or inside a crate since that affected if I could use `crate rustc -Zno-trans` or `rustc -Zno-trans`

Fixes #67

### DIFF
--- a/syntax_checkers/rust/rustc.vim
+++ b/syntax_checkers/rust/rustc.vim
@@ -14,9 +14,20 @@ let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_rust_rustc_GetLocList() dict
-    let makeprg = self.makeprgBuild({
-        \ 'args': 'rustc -Zno-trans',
-        \ 'fname': '' })
+    let cwd = '.' " Don't change cwd as default
+    let cargo_toml_path = findfile('Cargo.toml', '.;')
+    if empty(cargo_toml_path) " Plain rs file, not a crate
+        let makeprg = self.makeprgBuild({
+            \ 'exe': 'rustc',
+            \ 'args': '-Zno-trans' })
+    else " We are inside a crate
+        let makeprg = self.makeprgBuild({
+            \ 'exe': 'cargo',
+            \ 'args': 'rustc -Zno-trans',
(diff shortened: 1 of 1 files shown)


chris-morgan:
With anything like this I’m concerned about performance. Even no-trans can take many seconds for larger crates, where parse-only of the current file is universally fast.

faern:
You are right, this is a problem. On the other hand the linting really need to get better IMO. Do you know if there is any way to run this in the background that will not block the editor?

somini:
I'm sorry if I'm intruding, but this interests me, as syntax-only linting is pretty poor as @faern said.
I think the best course of action is making the fast behaviour the default, and having this as an option. You can then use syntastic in passive mode for rust files, meaning it will only run on `SyntasticCheck`. That should diminish the performance problems.

faern:
@somini Not intruding at all. You could continue on this PR if you would like to. I don't have any time to continue on this as I'm not using vim for Rust any longer.

ticki:
Any update on this?

faern:
Nope. I don't use vim any longer so I won't contribute to this. Could close the PR, but not sure if anyone else wanted to continue.

somini:
I'm tracking this on #76, so you can close this.

xftroxgpx:
@faern what do you use instead? curious if I would be interested to make the same change :)

PR #83, opened 2016-05-20
Added errorformat line: ignore 'Compiling...'

nyarly:
The motivation here was that `:make build` didn't jump immediately to the first error because of Cargo's "Compiling <file>..." line. Ignoring that takes us right to what needs fixing.


nyarly:
@chris-morgan Any input on this?

steveklabnik:
Thanks!

PR #93, opened 2016-07-15
Fix incorrect indent on new match arms

C4K3:
This adds another case where if the previous line ends on a comma and
the current line is a new match arm, then we *don't* use the previous
line's indentation.

For example consider

```
match x {
    a => f(0,
           1),
    b => ...
}
```

which would previously be indented as

```
match x {
    a => f(0,
           1),
           b => ...
}
```

### DIFF
--- a/indent/rust.vim
+++ b/indent/rust.vim
@@ -1,7 +1,7 @@
 " Vim indent file
 " Language:         Rust
 " Author:           Chris Morgan <me@chrismorgan.info>
-" Last Change:      2014 Sep 13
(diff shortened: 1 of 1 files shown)


chris-morgan:
Thanks; seems good. Of course, all of this stuff is fundamentally flawed, but it’s good to get it closer to accurate where it’s possible to do so. It’d be really nice if there were something like racer that we could pass this off to and get its opinion of what indentation should be suggested at such-and-such a point.

PR #94, opened 2016-07-27
Move rustfmt_autosave to ftplugin

michael-younkin:
It doesn't need to be in the `plugin` folder because the autocommand only needs to be set when a Rust file is loaded. Only things that need to run when Vim starts should go in the `plugin` folder. Everything else can be put in `ftplugin`.

Behavior should be the same except that it picks up the value of `g:rustfmt_autosave` later than it did before. This is how most other plugins behave.

### DIFF
--- a/ftplugin/rust.vim
+++ b/ftplugin/rust.vim
@@ -193,6 +193,11 @@ let b:undo_ftplugin = "
 
 " }}}1
 
+" Code formatting on save
+if get(g:, "rustfmt_autosave", 0)
+	autocmd BufWritePre *.rs call rustfmt#Format()
+endif
+
 augroup END
 
 let &cpo = s:save_cpo
--- a/plugin/rustfmt.vim
+++ b/plugin/rustfmt.vim
@@ -1,8 +1,0 @@
-augroup rustfmt
-  autocmd!
-
-  " code formatting on save
-  if get(g:, "rustfmt_autosave", 0)
-    autocmd BufWritePre *.rs call rustfmt#Format()
-  endif
-augroup END


steveklabnik:
> Only things that need to run when Vim starts should go in the plugin folder. Everything else can be put in ftplugin.

Can you provide a citation for this? That's not my understanding. `ftplugin` is for filetype-specific stuff, and `plugin` is for more general things.

michael-younkin:
Right, `ftplugin` is for filetype-specific stuff, and `rustfmt_autosave` is only relevant for the Rust filetype. The autocommand only needs to be set if you're going to be editing Rust code. Currently, however, it gets set every time Vim starts, which is unnecessary.

I know it's not really a big performance issue, but it is more correct, and at the very least it simplifies the codebase a bit.

The actual vim docs are a bit vague on when the different folders get loaded, but [_Learn Vimscript the Hard Way_](http://learnvimscriptthehardway.stevelosh.com/chapters/42.html) explains it well: "Files inside `~/.vim/plugin/` will each be run once _every time_ Vim starts. These files are meant to contain code that you always want loaded whenever you start Vim." and "Files in `~/.vim/ftplugin/` are different. ... For example: if you run `set filetype=derp` Vim will look for `~/.vim/ftplugin/derp.vim`. If that file exists, it will run it."

steveklabnik:
> I know it's not really a big performance issue, but it is more correct, and at the very least it simplifies the codebase a bit.

Cool, I agree, I just saw it as the wrong justification. Let's :shipit:

Issue #88, opened 2016-06-09
Error message update RFC and IDE impact


jonathandturner:
Hey all,

Sorry for using an issue as a message box, but I wasn't sure who to contact.

I/we are currently working on an RFC that will change up the compiler error output a bit: https://github.com/rust-lang/rfcs/pull/1644  This may have some IDE impacts, if the plugin is doing a match against the output.

We've also been talking about standardizing a JSON output for the compiler so that IDEs can get more information that way.

Would love to get your feedback on what impact it has on the vim plugin.

chris-morgan:
An issue is the right way of handling it! (BTW, there’s the @rust-lang/rust-vim team so you can call us (OK, only me and brson at present) in on an issue in another repository if you want.)

Vim’s compiler error handling is text-based; see the [`errorformat` docs](http://vimdoc.sourceforge.net/htmldoc/quickfix.html#errorformat) for info. It also supports only one line of message per error, which probably means that we’re not going to be able to transmit *everything*. It will necessarily be a slightly more limited thing. For multi-line messages you could craft multiple error lines, each with a part of the full message, but that’s not the nicest thing (though it may be helpful) and it does mess up the error counts.

If the compiler were to emit exclusively JSON, we’d need a way of parsing it. For `'makeprg'`, I don’t believe there’s a way of processing the output of the program before errorformats has at it. That means writing a program wrapping rustc/cargo and turning its JSON into something simple to parse. (That means either making a slow mess of shell/batch + vimscripts for JSON parsing, or writing something in Rust, which is guaranteed to be available.) I’d certainly prefer it if rustc had two forms of output for tooling: JSON with all the info, and plain text single-line-per-message mostly as it used to be (e.g. `file.rs:1:1: error: message [E0000]`, but making sure it’s only one line per message) form.

I’ve ignored syntastic in this; it’s going to be equal to or more powerful than errorformat/makeprg. (There may be a way of hooking a decoding function into it that isn’t string-based; I don’t know.)

nikomatsakis:
@chris-morgan 

> It also supports only one line of message per error, which probably means that we’re not going to be able to transmit everything.

I'm not quite sure what this means -- there is a whole section on matching multi-line error messages?

chris-morgan:
@nikomatsakis I mean that it supports only one line of descriptive text per error message. So if you have this:

    file:line:col error: such and such
        and also so and so

You can’t get the likes of “`such and such\nand also so and so`” as the message of an error.

[...]

PR #99, opened 2016-08-15
Add support for new error format

jpernst:
This adds basic support for the new rustc error format used in recent nightlies. It appears to be working, but it's possible I've missed some edge cases, so comments are welcome.

Also, I'm not at all familiar with syntastic or how it works, so I haven't tested that part yet. If someone who uses that plugin could give this a try and let me know if it's broken, that would be great.

### DIFF
--- a/compiler/cargo.vim
+++ b/compiler/cargo.vim
@@ -27,7 +27,10 @@ function! s:is_absolute(path)
     return a:path[0] == '/' || a:path =~ '[A-Z]\+:'
 endfunction
 
-CompilerSet errorformat+=%-G%\\s%#Compiling%.%#
+" Ignore general cargo progress messages
+CompilerSet errorformat+=
+			\%-G%\\s%#Downloading%.%#,
+			\%-G%\\s%#Compiling%.%#
 
 let s:local_manifest = findfile(s:cargo_manifest_name, '.;')
 if s:local_manifest != ''
--- a/compiler/cargo.vim
+++ b/compiler/cargo.vim
@@ -30,7 +30,8 @@ endfunction
 " Ignore general cargo progress messages
 CompilerSet errorformat+=
 			\%-G%\\s%#Downloading%.%#,
-			\%-G%\\s%#Compiling%.%#
+			\%-G%\\s%#Compiling%.%#,
+			\%-G%\\s%#Finished%.%#
 
 let s:local_manifest = findfile(s:cargo_manifest_name, '.;')
 if s:local_manifest != ''
--- a/compiler/cargo.vim
+++ b/compiler/cargo.vim
@@ -31,7 +31,9 @@ endfunction
 CompilerSet errorformat+=
 			\%-G%\\s%#Downloading%.%#,
 			\%-G%\\s%#Compiling%.%#,
-			\%-G%\\s%#Finished%.%#
+			\%-G%\\s%#Finished%.%#,
+			\%-G%\\s%#error:\ Could\ not\ compile\ %.%#,
+			\%-G%\\s%#To\ learn\ more\\,%.%#
 
 let s:local_manifest = findfile(s:cargo_manifest_name, '.;')
 if s:local_manifest != ''
--- a/compiler/rustc.vim
+++ b/compiler/rustc.vim
@@ -39,8 +39,7 @@ CompilerSet errorformat+=
 			\%Eerror[E%n]:\ %m,
 			\%Wwarning:\ %m,
 			\%Inote:\ %m,
-			\%C\ %#-->\ %f:%l:%c,
-			\%C%*[0-9\ ]\|%.%#
+			\%C\ %#-->\ %f:%l:%c
 
 let &cpo = s:cpo_save
 unlet s:cpo_save
--- a/compiler/rustc.vim
+++ b/compiler/rustc.vim
@@ -21,6 +21,7 @@ else
 	CompilerSet makeprg=rustc\ \%
 endif
 
+" Old errorformat (before nightly 2016/08/10)
 CompilerSet errorformat=
 			\%f:%l:%c:\ %t%*[^:]:\ %m,
 			\%f:%l:%c:\ %*\\d:%*\\d\ %t%*[^:]:\ %m,
@@ -28,6 +29,15 @@ CompilerSet errorformat=
 			\%-G%*[\ ]^,
 			\%-G%*[\ ]^%*[~],
 			\%-G%*[\ ]...
+
+" New errorformat (after nightly 2016/08/10)
+CompilerSet errorformat+=
+			\%Eerror:\ %m,
+			\%Eerror[E%n]:\ %m,
+			\%Wwarning:\ %m,
+			\%Inote:\ %m,
+			\%C\ %#-->\ %f:%l:%c,
+			\%C%.%#
 
 let &cpo = s:cpo_save
 unlet s:cpo_save
--- a/compiler/rustc.vim
+++ b/compiler/rustc.vim
@@ -32,12 +32,15 @@ CompilerSet errorformat=
 
 " New errorformat (after nightly 2016/08/10)
 CompilerSet errorformat+=
+			\%-G,
+			\%-Gerror:\ aborting\ %.%#,
+			\%-Gerror:\ Could\ not\ compile\ %.%#,
 			\%Eerror:\ %m,
 			\%Eerror[E%n]:\ %m,
 			\%Wwarning:\ %m,
 			\%Inote:\ %m,
 			\%C\ %#-->\ %f:%l:%c,
-			\%C%.%#
+			\%C%*[0-9\ ]\\|%.%#
 
 let &cpo = s:cpo_save
 unlet s:cpo_save
--- a/compiler/rustc.vim
+++ b/compiler/rustc.vim
@@ -40,7 +40,7 @@ CompilerSet errorformat+=
 			\%Wwarning:\ %m,
 			\%Inote:\ %m,
 			\%C\ %#-->\ %f:%l:%c,
-			\%C%*[0-9\ ]\\|%.%#
+			\%C%*[0-9\ ]\|%.%#
 
 let &cpo = s:cpo_save
 unlet s:cpo_save
--- a/syntax_checkers/rust/rustc.vim
+++ b/syntax_checkers/rust/rustc.vim
@@ -20,19 +20,16 @@ function! SyntaxCheckers_rust_rustc_GetLocList() dict
     let errorformat  =
         \ '%E%f:%l:%c: %\d%#:%\d%# %.%\{-}error:%.%\{-} %m,'   .
         \ '%W%f:%l:%c: %\d%#:%\d%# %.%\{-}warning:%.%\{-} %m,' .
-        \ '%C%f:%l %m,' .
-        \ '%-Z%.%#'
+        \ '%C%f:%l %m,'
         
     " New errorformat (after nightly 2016/08/10)
     let errorformat  .=
-        \ ',' .
         \ '%Eerror: %m,' .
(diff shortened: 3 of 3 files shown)


ExpHP:
> Also, I'm not at all familiar with syntastic or how it works, so I haven't tested that part yet. If someone who uses that plugin could give this a try and let me know if it's broken, that would be great.

It seems the pieces are not all quite there yet.  I don't know too much about it all myself, so here is an image comparison of how syntastic looks in both the current version and your PR, and on stable versus nightly:

http://imgur.com/a/5j4xp

jpernst:
Ok, I think the reason for this is that the old errorformat for syntastic had an eager match that prevented the subsequent lines of the new format from matching. I've removed that line; see if this works better now.

ExpHP:
I added two new images [to the album](http://imgur.com/a/5j4xp).  The new error format messages are now properly integrated into the interface (I can write code again!! :tada:)  However there is a discrepancy in where matching ends against the old format error messages.

jpernst:
Yeah, that extra error message is a bit vexing for me as well. Unfortunately, with how the new error format works, it's completely indistinguishable from a "real" error, and there are several such "noise" messages. I can't think of a good way to deal with this without resorting to lots of hardcoded cases to ignore them all individually.

If someone has a better way I'd be interested to know.

ExpHP:
I recall the PR which added the new error messages [also says something about a "json mode"](https://github.com/rust-lang/rust/pull/35401)).  Rummaging around the options it looks one can use `rustc --error-format json` (adding `-Z unstable-options` on stable).

It's certainly a easier to read format, though I wouldn't dare try parsing it with a regex... I wonder if there is another way to configure these plugins without regex?

jpernst:
I mentioned this on the internals board, and they seem to indicate that those extra messages aren't final and may be revised, so I'm inclined to just live with them until the dust settles, then we'll see.

kesselborn:
Hey, I just implemented the (I think) same thing today, but perhaps it's not exactly the same? I wanted to have proper quickfix browsing in the following cases:

  - `cargo build` reports errors
  - `cargo build` reports warnings
  - `cargo test` reports errors

If I try out the current pull request, this does not work for me ... or is this PR only for syntastic (which I don't really use) only?

Adding the following snippet to `ftplugin/rust.vim` works for me correctly with the old and the new error formats (nightly and stable on September 4th, 2016):

```vim
" order matters:
" newer rustc errors and warnings
setl errorformat=%Eerror%m,%Z\ \ -->\ %f:%l:%c,%Wwarning%m,%Z\ \ \ -->\ %f:%l:%c
" older rustc errors and warnings
setl errorformat+=%f:%l:%c:\ %*[0-9]:%*[0-9]\ %trror:%m,%f:%l:%c:\ %*[0-9]:%*[0-9]\ %tarning:%m
" cargo test error messages:
setl errorformat+=%E----%m,%Z%m\ %f:%l
```

jpernst:
This PR should work (and does for me) with cargo; I'm not sure why it isn't for you. I get errors and warnings from cargo just fine, except for "test" which I haven't added, since I don't usually run that from inside vim.

MarkSwanson:
kesselborn, your little snippet works great for me (but not quite perfect). Scenarios I tried (via vim :mak):
(I use vim-localvimrc to set makeprog to use a shell script to exec cargo)

- syntax error in current src/file.rs -> takes me to the correct line number. Great!

- syntax error in different src/file2.rs -> takes me to the correct line number in the other file. Great!

- *test* syntax error in current test/file.rs -> fail. Does not move the cursor to the error line reported by cargo.

kesselborn, my test output is rather large and might be breaking things for you. Here is a snippet of what happens when I try to compile test/file.rs on my machine:

```
Compiling file v1.0.4 (file:///home/maswanso/src/file/test/file.rs)
     Running `rustc file.rs --crate-name file -g --test -C metadata=2ac560e36b429231 <800 characters...>
error: expected item, found `ufse`
  --> file.rs:34:1
   |
34 | ufse file_logger::spdlog_wrapper::{Spdlog, SpdlogProxy, FileLogger};

error: aborting due to previous error

error: Could not compile `file`.

Caused by:
  Process didn't exit successfully: `rustc file.rs <800 characters...>

```

jpernst - I'm sure it works fine for you on your machine. :-) That's just how it goes sometimes :-) But your patches don't yet handle the new format on my machine.

MarkSwanson:
Er... it's late and I must be tired because it's no longer taking me to the right file or line number.

I've opened up the vim help on errorformat and am playing around. I note the example Python multiline format string uses \C. Playing...

MarkSwanson:
Ok, I've played around with things a bit. Here's what works for me 100% in every case:

```
setl errorformat=%Eerror:\ %m,%Z\ \ -->\ %f:%l:%c
```

I do have 'error' in the   ~11000 (eleven thousand) characters that vim has to regex search through. I thought adding the ':\ ' might help but I don't think it did because when I start putting things back like this:

```
setl errorformat=%Eerror:\ %m,%Z\ \ -->\ %f:%l:%c,%Wwarning%m,%Z\ \ \ -->\ %f:%l:%c
```

Then things don't work anymore.

I'm a bit lost in the weeds I think :-)
Hope this helps. If you want me to try stuff please let me know.

Thanks!

MarkSwanson:
Ok, I circled back and noticed a bug in the Rust code that generates the error.

Bug: It 'non-deterministically' uses 2 or 3 spaces in front of "-->"

```
error[E0425]: unresolved name `MY_ERROR_V1`
   --> src/server.rs:111:11
```

This is why it 'stopped' working for me earlier - different errors have different spaces.

What seems to work for me:

```
setl errorformat=\ \ \ -->\ %f:%l:%c,\ \ -->\ %f:%l:%c
```

Obviously it's a hack for now - but I'm still figuring out the different regex syntax (to support 0-n spaces before '-->') and all of the vim errorformat mechanisms...

jpernst:
One possibility for why it's not working for some: In the original plugin (and in my PR), the `errorformat` is set for the `cargo` compiler, which must explicitly be enabled when editing to take effect. In my `init.vim` I have an autocmd that enables it as follows:
`autocmd FileType rust compiler cargo`
If cargo is being invoked some other way, the proper `errorformat` might not be present. Just a hunch anyway.

As for detecting the padding, `%#` after the space should have the effect of a regex `*` and capture any number of spaces. I believe the arrow's offset is determined by the number of digits in the linenumber, but that's just what I've observed, and can't say for sure if that always the case.

MarkSwanson:
Ok, the %# worked like a charm. Thanks!

```
setl errorformat=%Eerror%m,%Z\ %#-->\ %f:%l:%c
```

Interesting hunch about the autocmd. I actually don't set 'compiler'.
I've re-skimmed the docs on compiler a couple of times and still don't understand how that interacts with makeprg.

I submit to you that there is tremendous value in allowing people to have a rust compiler other than cargo. Using localvimrc enables me to dynamically change the makeprg. Examples:
- editing test/file.rs => I can run some variant of 'cargo test' against specific Fns with custom settings that are tuned to specific crates/files. I don't have to leave vim.
- I can do things like 'sudo' before running the test - and I'm prompted within vim for the root password. I don't have to leave vim.
- I'm doing other custom things like dynamically setting LD_LIBRARY_PATH or setting other variables.
- Various other things to make life easier.

Is it possible to just use:

``` 
autocmd FileType rust
```
(without the 'compiler cargo' qualification) ?

Thanks!

jpernst:
@MarkSwanson `autocmd FileType rust` just means "run the following command when you enter a buffer with a rust file", in this case the command being `compiler cargo`. It should be noted that the cargo compiler inherits settings from the rustc compiler, so the errorformat should be set properly for both. For other compiler configurations, the errorformat could simply be copied into them. Supporting rustc and cargo will probably cover 99% of most out-of-box uses. Adding different compiler configurations is probably something best discussed in another issue/PR.

@euclio Huh, I've never even used rustfmt before so I have no idea what's going on there. Perhaps I'll take a look, but if someone else is more familiar with rustfmt and what's supposed to happen wrt errors there, please chime in.

euclio:
Maybe @grncdr can comment since he wrote that code.

Either way, a workaround for now is to set `g:rustfmt_fail_silently=1`.

grncdr:
to say I wrote it is a bit of an exaggeration, I only hacked the vim-go plugin with some regexes that seemed to work with the toolchain I was using at the time. (it looks like vim-go [hasn't changed much since then](https://github.com/fatih/vim-go/blob/8575d9e3c9e23508f9b7d0c3992cb683b1e47ae6/autoload/go/fmt.vim#L171-L190))

IMO the most sensible approach is using `rustc` (with `--error-format json`) to check for syntax errors before running `rustfmt`. I don't know for sure what errors `rustfmt` might emit after that, but those are the only ones the line of code in question should be handling.

steveklabnik:
Well, the new errors are out, and I'm a bad maintainer. What's the state of this PR?

jpernst:
I just updated this yesterday to show all `note` lines, since they tend to be important, and to filter out various extraneous status messages from rustc/cargo. It's working well for me so far, but more testing is always welcome, particularly from people who use the syntastic support.

steveklabnik:
I am inclined to merge this, then. It seems small enough.

Others in this thread, what do you think?

ExpHP:
It has been looking pretty good on nightly.

I just dropped into a stable crate to test and [this bit of noise about -Z still shows up](http://i.imgur.com/dXHLnuS.png).

(on a tangential but closely related note, it might be a good idea to start pushing for stabilization of `-Zparse-only`? :stuck_out_tongue: )

jpernst:
Hmm, I forgot about that. I'm a little conflicted about filtering that since it indicates that the syntastic plugin will eventually break, but perhap's that's a problem for another day. I'll look into silencing it.

ExpHP:
I do rather feel the same way!  Perhaps keeping it around might make people a little more motivated... ;)

jpernst:
@ExpHP You make a good point; for now I'll leave that warning in. Also, I pushed some minor adjustments to the syntastic matcher, could you give it a try when you get a chance to make sure everything is still sane? If so, then this should be ready to merge.

ExpHP:
Still looks good to me (or at least, the parts I use)!

I did some comparisons trying to find what changed, and I see that the new commit filters out a bit of noise that I've apparently become blind to (the `|| aborting due to previous error`), so that's good too. :)

jpernst:
Yeah, having "fake" errors in the list can really mess things up, especially when it tries to jump to the error with no associated location, so I decided it was imporant to remove those from the list finally.

Sounds like things are pretty solid now, so unless someone else comes forward with a bug, I think we're good to go.

jpernst:
After using this for a little while longer, I've been forced to unhide all of the source context in the error messages. This adds back a significant amount of noise, but unfortunately it sometimes contains critical information that isn't otherwise shown, such as the exact types in a type mismatch. Lacking that information hurts more than sifting through the extra noise.

mbrubeck:
cc #88

steveklabnik:
Okay! It's been a day. Let's :shipit: so that it doesn't get lost again. We can always tweak things if they need fixing.

Thanks all!

irfansharif:
@ExpHP: newbie to rust and rust.vim here, concerning the `option Z is unstable..` noise, how do i get rid of it?

withoutboats:
I'd really like to see the unstable flag error filtered out of Syntastic's error list. Exposing this to rust.vim users is very frustrating. It is certainly not end users' responsibility to be concerned about the stability of this plugin's integration with rustc.

Rolandde:
@withoutboats : Another way to silence the message is to use Synstatic's option
     let g:syntastic_quiet_messages = {"regex": 'is unstable and should only be used on the nightly compiler, but it is currently accepted for backwards compatibility; this will soon change, see issue #31847 for more details'}

I'm not sure how this affects speed, but I doubt it will match any other errors.

PR #107, opened 2016-09-21
Fail silently on format failure

kopiczko:
Before this change when a file with syntax error was written, all its
content was printed as a formatting error message. This change makes
format command fail silently and prevents form such behaviour.


steveklabnik:
Makes sense to me, thank you!

PR #116, opened 2016-10-18
Don't fix paths after running cargo.

stoklund:
Recently, vim started reporting Rust compilation errors in non-existent filenames in my multi-crate repository. Possibly after I installed rustc 1.12. The problem is caused by the FixPaths function in compiler/cargo.vim which would rewrite `Cargo.toml`-relative paths to current directory-relative paths. However, it would use the wrong `Cargo.toml`.

Cargo now emits paths that are relative to the current directory, not `Cargo.toml`, so the path rewriting is no longer necessary.

Also remove the documentation for the `g:cargo_manifest_name` setting which isn't used any more.

### DIFF
--- a/compiler/cargo.vim
+++ b/compiler/cargo.vim
@@ -19,14 +19,6 @@ else
     CompilerSet makeprg=cargo\ $*
 endif
 
-" Allow a configurable global Cargo.toml name. This makes it easy to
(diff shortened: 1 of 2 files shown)


steveklabnik:
Great, thanks!

PR #117, opened 2016-10-22
added UltiSnips

roberbnd:
Hallo
I used the code of vim-go
https://github.com/fatih/vim-go/blob/master/ftplugin/go/snippets.vim
I wonder if it possible add snippets of ultisnips to create the rest of snippets.

### DIFF
--- a/ftplugin/rust/snippets.vim
+++ b/ftplugin/rust/snippets.vim
@@ -1,0 +1,46 @@
+if exists("g:rustg_loaded_rustsnippets")
+  finish
+endif
+let g:rust_loaded_rustsnippets = 1
+
+" by default UltiSnips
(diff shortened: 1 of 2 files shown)


steveklabnik:
I'm not sure about this patch; it's not clear that this is something we'd want for everyone...

roberbnd:
just who use Ultisnippts, its ok we can wait if someone is agree or not

steveklabnik:
It seems that nobody else desires this a ton, so I'm going to close it. Thanks!

onionjake:
I wouldn't mind this (sorry so late).

This repo also has some rust ultisnips: https://github.com/honza/vim-snippets/blob/master/UltiSnips/rust.snippets  @roberbnd maybe that would be a better spot?

PR #121, opened 2016-11-10
compiler/cargo.vim: remove FixPaths() function

billiob:
With cargo from rust 1.12.1 (but behaviour might have changed in earlier
versions), cargo is calling rustc with path related to the current
directory. There is no longer need to call FixPaths()
I had issue with FixPath because I'm calling cargo build from the src/
directory but FixPath would show errors on paths that do no contain src/.

### DIFF
--- a/compiler/cargo.vim
+++ b/compiler/cargo.vim
@@ -23,10 +23,6 @@ endif
 " support variations like 'cargo.toml'.
 let s:cargo_manifest_name = get(g:, 'cargo_manifest_name', 'Cargo.toml')
 
-function! s:is_absolute(path)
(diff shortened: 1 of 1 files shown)


jpernst:
I've been having intermittent problems with quickfix errors opening incorrect files or files that don't even exist, particularly when working on a project with nested crates. Took a while to finally track down the cause, and `FixPaths` was the culprit. I've applied this patch and the issue is resolved, so +1 from me.

steveklabnik:
I've merged https://github.com/rust-lang/rust.vim/pull/116 which seems to do much the same thing. What do you think, @billiob ?

billiob:
Looks fine.

Issue #120, opened 2016-10-29
Errors when loading .rs files after commit eb716cb


iirelu:
Sample error:

```
"src/libsyntax/parse/mod.rs" 1097L, 43180C
Error detected while processing function <SNR>13_LoadFTPlugin:
line    2:
E216: No such group or event: rust.vim|augroup! rust.vim
Press ENTER or type command to continue
```

This happens on every load of a Rust file. No errors before eb716cb.

tarrant:
Hmm, I had zero issues with this when I wrote https://github.com/rust-lang/rust.vim/commit/eb716cba6b8a6b8e156cdee0018c92a29a5a78d8 on my other computer. Just upgraded on this computer and I'm seeing the same issue. Hopefully I can narrow down why this is happening shortly. Apologies for introducing this bug!

tarrant:
It looks like the `augroup` stuff just really struggles being in a multi-line command. It could be wrapped in a function and then the function could be deleted, or it could just be left alone and not included in the post-buffer cleanup. Which should be fine since the autocommands are defined on a `*.rs` file only basis.

Thoughts opinions?

tarrant:
So the more I've thought about it the more it seems that removing the `augroup` stuff from the `b:undo_ftplugin` is the right way to go. These are already correctly cleaned up via other methods and there doesn't seem to be a clean way (that I can find) to deal with them here.

I'm hesitant to fix this without some feedback since I'm the one that broke it in the first place :( but if no one responds I'll submit a patch anyways.

tmccombs:
This happens to me in neovim, with slightly different message:

> Error detected while processing function <SNR>7_LoadFTPlugin:
> line    2:
> E216: No such group or event: rust.vim|augroup! rust.vim

k-mack:
I don't get these errors when I load a Rust file, but I get something similar when I run `:RustFmt`
> Error detected while processing function <SNR>10_LoadFTPlugin:
> line    2:
> E216: No such group or event: rust.vim|augroup! rust.vim

tarrant:
I have in theory a fix on my fork at [tarrant/rust.vim](https://github.com/tarrant/rust.vim). Mind giving it a try and seeing if it resolves your issues? The more people that test it the better I'll feel about opening a PR.

k-mack:
@tarrant your fork doesn't result in the errors I was seeing before, so that's good. BTW, I am using Vim 7.4.

GustavoKatel:
[...]

PR #124, opened 2016-11-11
Remove all augroup references from b:undo_ftplugin

tarrant:
This was causing additional issues on some setups of vim as show in
[issue#120](https://github.com/rust-lang/rust.vim/issues/120). I thought about several ways to resolve this and my albeit
limited knowledge of vimscript couldn't come up with one. Ultimately I
resolved that it shouldn't be an issue since the augroup already clears
itself every time the augroup loads itself. In addition the aucommands
are already limited in scope to `*.rs` buffers. So they should be safe.

From my testing the core issue seems to be that `autocmd!` (on some
versions of vim?) doesn't stop parsing the name of the augroup at the
end of the line so instead of it trying to clear `autocmd! rust.vim` it
instead attempts to clear `autocmd! rust.vim|augroup!rust.vim` which
obviously isn't the name of the `augroup`.

### DIFF
--- a/ftplugin/rust.vim
+++ b/ftplugin/rust.vim
@@ -177,8 +177,6 @@ let b:undo_ftplugin = "
 		\|ounmap <buffer> ]]
 		\|set matchpairs-=<:>
 		\|unlet b:match_skip
-		\|autocmd! rust.vim
-		\|augroup! rust.vim
 		\"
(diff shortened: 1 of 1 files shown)


steveklabnik:
Great! Thanks so much.

PR #126, opened 2016-11-15
add rustQuestionMark

seanmonstar:
The pattern matches these examples:

```rust
let x = Err("foo")?;
let y = {
     Err("bar")?
};
let z = foo()?.bar()?;
```

And does not match:

```rust
fn foo<T: ?Sized>(val: &T) {
}
```


steveklabnik:
Thanks!

chris-morgan:
I was leaning towards `Operator` rather than `Special`, myself.

PR #127, opened 2016-11-22
Avoid warning of -Z due to #31793

gywn:
from rust-lang/rust#31793 the usage of -Z on the stable channel will now emit a warning, which should be avoided in syntax checking.


steveklabnik:
Thank you!

Issue #113, opened 2016-10-10
The option -Z is now unstable


mkollaro:
I'm getting the following warning for all Rust files:
_The option `Z` is unstable and should only be used on the nightly compiler, but it is currently accepted for backwards compatibility; this will soon change, see issue [#31847](https://github.com/rust-lang/rust/issues/31847) for more details._

I'm using rustc 1.12.0 (3191fbae9 2016-09-23) and rust.vim ref 2030019.

```
Syntastic version: 3.6.0                                                                                                                 
Info for filetype: rust                                                                                                                  
Global mode: active                                                                                                                      
Filetype rust is active                                                                                                                  
The current file will be checked automatically                                                                                           
Available checker: rustc                                                                                                                 
Currently enabled checker: rustc
```

steveklabnik:
Hm, https://github.com/rust-lang/rust.vim/blob/fc11d02fee330df7b30c83a80f09dd0c60ab43ce/autoload/rust.vim#L102 is the spot that it's used inside the extension. The file hasn't changed in a long time...

spinda:
Is there any way to silence this warning for the time being? It's quite annoying having the Syntastic window always open for .rs files.

Rolandde:
@spinda

     let g:syntastic_quiet_messages = {"regex": 'is unstable and should only be used on the nightly compiler, but it is currently accepted for backwards compatibility; this will soon change, see issue #31847 for more details'}

I'm not sure how this affects speed, but I doubt it will match any other errors.

spinda:
Thank you!

lpenz:
I've tried adding

    let g:syntastic_quiet_messages = {"regex": 'is unstable and should only be used on the nightly compiler, but it is currently accepted for backwards compatibility; this will soon change, see issue #31847 for more details'}

It does suppress the message, but I don't get any warnings from syntastic. I was only able to get syntastic working again by changing the plugin, initializing `makeprg` as follows:

    let makeprg = self.makeprgBuild({})

BurntSushi:
I just had the chance to update my rust-vim package and I'm now experiencing this change. I'm not really sure how the existing configuration is supposed to work at all in most projects, since it's effectively just calling `rustc the_file.rs`. For example, such a command always fails on the first `extern crate` it sees, since `rustc` isn't told where to find the crate.

In order to make the syntastic plugin useful to me again, I added this to my `.vimrc`, which restores the original behavior:

```
[...]

PR #128, opened 2016-11-27
Do not call rustc with -Z arguments in syntastic plugin

lpenz:
The flag is unstable, and is now generating a warning - which makes
the syntastic plugin unusable. Without it, we probably have a longer
analysis time, but it at least works. Issue rust-lang/rust#31847
explains the rustc side.

This fixes rust-lang/rust.vim#113.


steveklabnik:
Thanks!

PR #129, opened 2016-11-29
added cargo command

roberbnd:
The cargo command 
:Cargo
Cbuild
Cclean
Cdoc
Cnew -> creates the project and goes inside
Cinit
Crun
Ctest
Cbench
Cupdate
Csearch  
Cpublish 
Cinstall

### DIFF
--- a/autoload/cargo.vim
+++ b/autoload/cargo.vim
@@ -1,0 +1,85 @@
+function! cargo#cmd(args)
+    silent! clear
+    if !a:args
+        execute "!" . "cargo ". a:args
+    else
+        echom "Missing arguments"
+    endif
+endfunction
+
+function! cargo#build(args)
+    silent! clear
+    if !a:args
+        execute "!" . "cargo build " . a:args
+    else
+        execute "!" . "cargo build"
+    endif
+    silent! clear
+    execute "!" . "cargo build"
+endfunction
+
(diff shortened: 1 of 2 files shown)


steveklabnik:
because i'm terrible at vimscript, what does this do that https://github.com/rust-lang/rust.vim/blob/master/compiler/cargo.vim doesn't already give us?

roberbnd:
be supposed to  is use to compile the code but I never I found where is use it, if you want to use the folder "compiler"  is with the command :compiler and I  can't found it. The compilation is with 
`
let output = s:system(pwd, shellescape(rustc) . " " . join(map(rustc_args, 'shellescape(v:val)')))
`
line 69 of autoload/rust.vim
another hand my idea is create a interface to cargo like [fugiteve](https://github.com/tpope/vim-fugitive) to Git
now you can execute any command of cargo without leave vim, now is basic (you can supply with :!cargo args ) but in the future we can add more features

grawlinson:
I would love if this made it into rust.vim.

What's delaying the merge?

steveklabnik:
> What's delaying the merge?

https://github.com/rust-lang/rust.vim/issues/148 is delaying everything at the moment.

steveklabnik:
Let's do this

Sorry about the wait.

PR #133, opened 2017-01-01
Support i128 and u128

wdv4758h:
Rust now have `i128` and `u128` in nightly, behind the feature flag.
We'll need to highlight them in Vim 😃

Ref: [i128 and u128 support](https://github.com/rust-lang/rust/pull/38482)


chris-morgan:
Thanks; all bases covered, in it goes.

PR #135, opened 2017-01-05
Add support for highlighting white space errors.

jrmuizel:
This is modeled after c_space_errors and java_space_errors

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -209,6 +209,16 @@ syn keyword rustTodo contained TODO FIXME XXX NB NOTE
 " FIXME: use the AST to make really good folding
 syn region rustFoldBraces start="{" end="}" transparent fold
 
+" When wanted, highlight trailing white space
+if exists("rust_space_errors")
+  if !exists("rust_no_trail_space_error")
+    syn match   rustSpaceError  "\s\+$"
+  endif
+  if !exists("rust_no_tab_space_error")
+    syn match   rustSpaceError  " \+\t"me=e-1
+  endif
+endif
+
 " Default highlighting {{{1
 hi def link rustDecNumber       rustNumber
 hi def link rustHexNumber       rustNumber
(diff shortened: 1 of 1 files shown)


chris-morgan:
This needs documentation before it can be accepted.

If we’re rebelling against spaces followed by tabs, might we not just as well rebel against tabs altogether?

jrmuizel:
I've added a comment describing what it does. Is that sufficient?

As for rebelling against tabs all together, I think it's worth matching the semantics of c_space_errors and java_space_errors instead of trying to go above and beyond.

jrmuizel:
Ping?

steveklabnik:
Sorry @jrmuizel ; I haven't been as communicative as I'd like to be here.

I ended up sending `rust.vim` upstream, but upstream wanted a lot of changes. I made those changes, and got it landed, but have not yet re-integrated those changes here. I wanted to do that before merging more PRs and making a bigger mess of things. I haven't had time yet. Sorry :(

jacwah:
Highlighting trailing whitespace is definitely useful. As for the tabs and spaces thing, I agree it's more useful to do what C and Java are doing, and not surprise anyone, than come up with some "intelligence" of our own.

I think the reason tabs are allowed before spaces is the way some old C projects do indentation. I've seen projects use `tabstop=8 softtabstop=2 shiftwidth=2 noexpandtab`, which means 2 space indents, where every indent of 8 spaces is replaced by a tab.

screen shot 2017-08-01 at 12 32 02

da-x:
Thanks for you contribution!

However I think that it is better not to duplicate over the work of external plugins such as [ntpeters/vim-better-whitespace](https://github.com/ntpeters/vim-better-whitespace) and [bronson/vim-trailing-whitespace](https://github.com/bronson/vim-trailing-whitespace)), which already provide more comprehensive functionality for managing whitespace.

PR #145, opened 2017-03-14
Change `unsafe` highlighting to `Exception`

### LINKED ISSUE #144: Suggestion: Option to highlight `unsafe` in bold
I'm not sure there is a good default highlighting that unsafe could be mapped to for this setting.

This is something I use in my Rust + Vim setup and I think that unsafe is not just another keyword, so it would be great to have this way of highlighting Rust spread.

chris-morgan:
It has a new group now, `rustUnsafeKeyword`, which allows the user to easily customise the highlighting if they desire. (Example: if you wanted to visually forbid `unsafe`, you could highlight it as `Error` by putting `hi link rustUnsafeKeyword Error` in your `.vimrc` or `.vim/after/syntax/rust.vim`.)

I also made this group link to `Exception` rather than `Keyword`. This is potentially mildly controversial. It’s common for `Exception` and `Keyword` to be highlighted the same, so many users won’t see this change. Still, it *is* a change, and one people may not like. We could back that part of it out easily enough. (This is also the reason I made this a PR rather than just pushing it directly to rust-lang/rust.vim.)

Closes #144.

/cc @bluss

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -32,7 +32,8 @@ syn keyword   rustKeyword     in impl let
 syn keyword   rustKeyword     pub nextgroup=rustPubScope skipwhite skipempty
 syn keyword   rustKeyword     return
 syn keyword   rustSuper       super
-syn keyword   rustKeyword     unsafe where
+syn keyword   rustKeyword     where
+syn keyword   rustUnsafeKeyword unsafe
 syn keyword   rustKeyword     use nextgroup=rustModPath skipwhite skipempty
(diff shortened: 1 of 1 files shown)


bluss:
Great! :+1: If it has a separate name then either way is fine with me.

vmx:
This is exactly what I'm after (coming from Emacs where `unsafe` really stands out). Why is this one not yet merged?

steveklabnik:
Seems fine, thanks!

Issue #108, opened 2016-09-24
None of the commands work when installed with Vundle.


pinkisemils:
I installed and updated to the latest version today, and none of the `:Rust*` commands seem to be defined. `:help :RustFmt` works, and syntax is being highlighted, but nothing else works.

I'm running a nightly compiler, rust.vim from commit `fc11d02fee330df7b30c83a80f09dd0c60ab43ce` and Vim 8.0.

ExpHP:
Just to provide a data point, I believe I am seeing the same thing? (however I use dein, not Vundle)

Like the OP, I find that `:help RustFmt` brings up help documentation, but `:RustFmt` or `:RustEmitIr` or etc. produces `E492: Not an editor command: RustEmitIr`.

To be clear, I've never tried these commands before, so I'm not sure what to expect; but neither the repo README nor the `:help` info suggest that there are any additional configuration steps required to enable these features.

I also tried wiping my plugins and reinstalling with commit b2a5429ccdf22d6fc8 (circa a year ago) and I still see the same results.

Various setup bits:
* Rust 1.12 installed through rustup v0.5.0
* VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Jun 16 2016 10:50:38) Included patches: 1-1689, all features except `-mouse_sysmouse -tag_any_white -footer -mzscheme -hangul_input -python -xfontset -ebcdic -xterm_save -sun_workshop -mouse_jsbterm`

pinkisemils:
Coming back to the issue, everything works if I open any other filetype first, and then open a rust source file, in my case with ctrl+p. 

Even still, `RustEmitIr` and `RustEmitAsm` don't work, as they can't locate crates I've imported from outside. `RustFmt` works, but only if the code compiles. And it seems to slow down exponentially, as time goes on. That might be an issue with `rustfmt` itself.

ExpHP:
Ah!  It is a good thing you mention filetypes, because I had completely forgotten this absolutely critical fact:

**I had at some point deliberately removed `filetype plugin on` from my vimrc.**

Sure enough, adding `filetype plugin on`, I find that `:RustFmt` becomes a usable command.

It sounds like for you, a plugin associated with that "other filetype" may be setting this option as well.

pinkisemils:
Nope, everything else that would seemingly rely on `filetype plugin on` just works. And it does autodetect the file type if I open up a rust file _later on_. It's weird, but since it does work for you, it might be my particular config that's weird.

[...]

PR #150, opened 2017-04-06
adding an issues/solutions section.

dorfsmay:
Run int the "Not an editor command: RustFmt" issue and it took me a long time to find the right solution.


dorfsmay:
See Issue #108

PR #152, opened 2017-05-03
Integrate changes from downstream vim

kejadlen:
As requested in https://www.reddit.com/r/rust/comments/690iup/is_rustlangrustvim_maintained_pull_requests/dh2snv8/. I *think* it's correct, as far as I can tell. Not sure what's up with `syntax/rust.vim`, though.

### DIFF
--- a/after/syntax/rust.vim
+++ b/after/syntax/rust.vim
@@ -1,34 +1,295 @@
-if !exists('g:rust_conceal') || g:rust_conceal == 0 || !has('conceal') || &enc != 'utf-8'
+" Vim syntax file
+" Language:     Rust
+" Maintainer:   Patrick Walton <pcwalton@mozilla.com>
+" Maintainer:   Ben Blum <bblum@cs.cmu.edu>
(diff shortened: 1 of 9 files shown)


chetgurevitch:
Could someone take a look at this?

steveklabnik:
Let's do it. Thank you.

chris-morgan:
There are two functional changes in this, and they both confuse and surprise me:

- `after/syntax/rust.vim`: … what? This is a mistake. If upstream really has it like that, we need to fix it.
- `<`/`>` %-matching (matchpairs) got removed. #160 covers this. We’ll need to investigate why this was done also.

kejadlen:
For reference, upstream [after/syntax/rust.vim](https://github.com/rust-lang/rust.vim/blob/master/after/syntax/rust.vim) and a [ftplugin/rust.vim](https://github.com/vim/vim/blob/master/runtime/ftplugin/rust.vim), which is where `matchpairs` got lost.

Might be easier to look at the [diff for this PR w/the whitespace elided](https://github.com/rust-lang/rust.vim/pull/152/files?w=1).

PR #153, opened 2017-05-13
Quickfix mode: Support assertion failure messages

gsauthof:
Such that you can quickly navigate between assertion failure
locations when you run `cargo test` (i.e. via `:make test`).

The user experience is then analogous to what you get with GCC
and common unittesting C++/C libraries like Catch/Boost/libcheck
when calling `:make check` or `:make test` from Vim.


chris-morgan:
Rebased and committed.

PR #163, opened 2017-07-04
Execute the rustfmt prewrite hook at all times.

dpc:
Currently `rustfmt` won't be run if the `g:rustfmt_autosave` setting was
enabled during the runtime, after the plugin and `rs` file were already
loaded.

This patch executes the hook unconditionally and makes the actual
decision to run `rustfmt` on every save.

### DIFF
--- a/ftplugin/rust.vim
+++ b/ftplugin/rust.vim
@@ -184,10 +184,14 @@ let b:undo_ftplugin = "
 
 " }}}1
 
+function! RustWritePreHook()
+	if get(g:, "rustfmt_autosave", 0)
+		 silent! call rustfmt#Format()
+	endif
+endfunction
+
 " Code formatting on save
-if get(g:, "rustfmt_autosave", 0)
-	autocmd BufWritePre *.rs silent! call rustfmt#Format()
-endif
+autocmd BufWritePre *.rs call RustWritePreHook()
 
 augroup END
 


dpc:
Note: VimL novice here.

chris-morgan:
I can’t think of any reason why you’d want to twiddle `g:rustfmt_autosave`; if we wanted it to be more flexible like that we’d make it use a `b:` flag to augment the `g:` flag. As it is, I don’t think it’s justified. If you think it is, then we can reopen this and change it to look for a `b:rustfmt_autosave`.

dpc:
@chris-morgan I have an autocmd that changes `g:rusftmt_autosave` to 1 if `rustfmt.toml` is present in the root of a project, and 0 otherwise. Idea is: if the project has `rustfmt.toml`, they probably would like to me to use `rustfmt` automatically., if not, I don't want to include formatting changes in my diffs. However without `g:rustfmt_autosave` working at runtime, it doesn't really work when I switch between projects...

chris-morgan:
@dpc That sounds like a useful mode of operation which we should include. We could introduce a `b:rustfmt_autosave` flag, and have another variable (maybe `g:rustfmt_autosave_if_config_present`, or maybe a sentinel value for `g:rustfmt_autosave` like `"if rustfmt.toml present"`) that sets it automatically.

dpc:
That would be even more awesome!

da-x:
@chris-morgan I've implemented your suggestion adding `g:rustfmt_autosave_if_config_present`, detecting `rustfmt.toml` and utilizing `b:rustfmt_autosave`.

dpc:
@da-x You're my favorite person on github right now. :D . Thank you again.

PR #175, opened 2017-09-24
Add reserved keyword `proc`

ordovicia:
This PR adds `proc` to `rustReservedKeyword`.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -3,7 +3,7 @@
 " Maintainer:   Patrick Walton <pcwalton@mozilla.com>
 " Maintainer:   Ben Blum <bblum@cs.cmu.edu>
 " Maintainer:   Chris Morgan <me@chrismorgan.info>
-" Last Change:  Feb 24, 2016
+" Last Change:  Sep 24, 2017
 " For bugs, patches and license go to https://github.com/rust-lang/rust.vim
 
 if version < 600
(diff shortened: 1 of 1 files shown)


ordovicia:
Ping?

ordovicia:
`yield` is already added to `rustKeyword` [here](https://github.com/rust-lang/rust.vim/blob/b6d88adcf9867aa69f4d20d45d49bb54979842a4/syntax/rust.vim#L36) as well as `macro` and `box`.

I'm not sure they should be registered as `rustReservedKeyword`.
They are in "reserved keywords" list in [book](https://doc.rust-lang.org/book/second-edition/appendix-01-keywords.html) but can be used in nightly (unless I'm misunderstanding).

pickfire:
@ordovicia It's true that it's nightly-only but it's experimental right? My fault that I didn't check deeper since the docs showed that it belong to "Keywords Reserved for Future Use".

ordovicia:
@pickfire Right. `yield`, `macro`, and `box` are all currently experimental (`proc` is not available even in nightly TMK).
@steveklabnik How do you think? Should these keywords belong to `rustReservedKeyword`?

ordovicia:
The `proc` keyword is going to be removed from the reserved keywords list (rust-lang/rust#49699).

ordovicia:
Closing this PR since the `proc` keyword has been removed.
IMHO other keywords e.g. `yield` should be discussed in a separate issue.

PR #176, opened 2017-09-27
Check tests, examples, and benchmarks with cargo

matthew-mcallister:
With no extra arguments, `cargo check` does not check any files under `tests`, `examples`, or `benches`, so the Cargo Syntastic checker cannot be used. Passing `--tests`, etc. to Cargo addresses this problem.

The slowness of running `cargo check` is slightly mitigated by omitting the flags unless the file is under one of said directories (as opposed to just doing `cargo check --tests --examples --benches` every time).  The checker is still not perfect, as `cargo check` never scans unit tests and Syntastic unpredictably fails when multiple buffers are open.

### DIFF
--- a/ftplugin/rust.vim
+++ b/ftplugin/rust.vim
@@ -101,6 +101,22 @@ if has('conceal') && exists('g:rust_conceal') && g:rust_conceal != 0
 	setlocal conceallevel=2
 endif
 
+" When checking with syntastic, pass --tests, --examples, etc. to cargo based
+" on the file location
+if exists('g:loaded_syntastic_plugin')
+	let cargodir = fnamemodify(findfile('Cargo.toml', '.;'), ':p:h')
+	if !empty(cargodir)
+		let abspath = fnamemodify(expand('%'), ':p')
+		if abspath =~ '^' . cargodir . '/tests'
+			let b:syntastic_rust_cargo_args = 'check --tests'
+		elseif abspath =~ '^' . cargodir . '/examples'
+			let b:syntastic_rust_cargo_args = 'check --examples'
+		elseif abspath =~ '^' . cargodir . '/benches'
+			let b:syntastic_rust_cargo_args = 'check --benches'
+		endif
+	endif
+endif
+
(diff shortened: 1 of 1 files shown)


ftplugin/rust.vim
```diff
@@ -101,6 +101,22 @@ if has('conceal') && exists('g:rust_conceal') && g:rust_conceal != 0
 	setlocal conceallevel=2
 endif
 
+" When checking with syntastic, pass --tests, --examples, etc. to cargo based
+" on the file location
+if exists('g:loaded_syntastic_plugin')
+    let cargodir = fnamemodify(findfile('Cargo.toml', '.;'), ':p:h')
+    if !empty(cargodir)
+        let abspath = fnamemodify(expand('%'), ':p')
+        if abspath =~ '^' . cargodir . '/tests'
```
chris-morgan: Don’t use regular expressions for this; magic characters will break it, Windows which uses backslashes for paths will break it. I’m not immediately sure what the best way of handling it will be because of the path separator issue. But you want to check “does it start with cargodir PATH_SEP tests PATH_SEP” and so forth.

I have just discovered that I genuinely don’t know of a good way to join paths in vimscript, nor can I find anyone else giving an answer. Perhaps check if the first character is a slash, otherwise use a backslash? (`expand()` and `glob()` have side-effects pertaining to magic characters, so they’re not suitable.)

matthew-mcallister: Thanks for the review; sorry it took me a while to notice.

Please excuse the regex dumbness; I don't think I'm very good at vimscript.

I agree with you about using `expand` and `glob`; those functions are also affected by `wildignorecase` and probably other settings which would have to be temporarily reset. I didn't think of a better way to get the path separator either, but your suggestion should at least work for Unix and Windows AFAIK. This works now with special characters in the path and it will probably work on Windows.

da-x:
Thanks for you contribution!

However I think it is preferred to let the user control this via configuration (global and buffer-local). I've done the necessary changes in master for this.

PR #185, opened 2017-12-07
Fix RustFmtRange

christopher-l:
`RustFmtRange` called `rustfmt` without a filename, so it never terminated.


da-x:
Thanks for you contribution! However the PR is no longer relevant - I've implemented this a bit differently in master.

PR #188, opened 2017-12-17
Preserve file permissions on rustfmt

joaodelgado:
Currently running `RustFmt` is resetting the file permissions.

`RustFmt` does the following steps:

1. create a temporary file
2. copy the contents of the current buffer to the temporary file
3. run `rustfmt` on the temporary file
4. replace the current file with the temporary one

The problem is in the first step since it is performing a simple `writefile`, it inherits the default file permissions for the directory.

After this pull request, the file permissions are copied from the source file to the temporary file after step 2:

1. create a temporary file
2. copy the contents of the current buffer to the temporary file
3. copy file permissions
4. run `rustfmt` on the temporary file
5. replace the current file with the temporary one


da-x:
Thanks for you contribution! I have implemented this differently - instead of copying the permissions, the original file is preserved but replaced with the new content.

PR #190, opened 2018-01-10
Better indent of closing parenthesis

AdnoC:
Makes indentation of closing parenthesis on their own line match better with what rustfmt will do.

Makes it indent like this:

```
some_func(
    really.long(|line| with.lots).of("stuff"),
    |other| long(|line| which.deserves).its("own").line
)
```

instead of this:

```
some_func(
    really.long(|line| with.lots).of("stuff"),
    |other| long(|line| which.deserves).its("own").line
    )
```

I've been coding in an environment where I can't call `rustfmt` from vim. Having the code look nicer from the get-go would be great.

### DIFF
--- a/indent/rust.vim
+++ b/indent/rust.vim
@@ -1,7 +1,7 @@
 " Vim indent file
 " Language:         Rust
 " Author:           Chris Morgan <me@chrismorgan.info>
-" Last Change:      2017 Mar 21
+" Last Change:      2018 Jan 10
 " For bugs, patches and license go to https://github.com/rust-lang/rust.vim
 
 " Only load this indent file when no other was loaded.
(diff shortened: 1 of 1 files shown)


da-x:
I remember that I also bumped into this issue a few times, and it annoyed.

`m1` indeed seems more consistent with recommended style (ping @nrc to verify).

This what the Vim docs say:
```
	mN    When N is non-zero, line up a line starting with a closing
	      parentheses with the first character of the line with the
	      matching opening parentheses.  (default 0).

		cino=(s			  cino=(s,m1 >
		  c = c1 && (		    c = c1 && (
		      c2 ||			c2 ||
		      c3			c3
		      ) && c4;		    ) && c4;
		  if (			    if (
		      c1 && c2			c1 && c2
		     )			    )
		      foo;			foo;
<
```

nrc:
Yes, the description in the OP moves things to what rustfmt would do.

PR #191, opened 2018-01-28
Cleanup cargo commands

rhysd:
I found some mistakes and redundancy in implementation of cargo commands. I fixed them as following:

- `:clear` is abbreviation of `:clearjump` and it is not necessary for running command
- `-nargs=+` checks the command should have one or more args. So it should not be checked in a function
- `"." . "foo"` is the same as `". foo"`.
- Some commands are wrongly run twice
- Remove boring boilerplates by reusing `cargo#cmd()`
- Fix the guard variable name. `plugin/rust.vim` is not only for syntastic
- Fix position of cpo-save in `plugin/rust.vim`. It must be restored at the end of the file

### DIFF
--- a/autoload/cargo.vim
+++ b/autoload/cargo.vim
@@ -8,7 +8,6 @@ function! cargo#build(args)
     else
         execute "! cargo build"
     endif
-    execute "! cargo build"
 endfunction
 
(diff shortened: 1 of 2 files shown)


da-x:
Thanks!

PR #201, opened 2018-03-27
Added mappings for cargo run, build, check

asoderman:
Not sure if this feature was ever requested but I added optional keybindings to run, build or check rust projects from within vim. I use this in my personal .vimrc and thought it would be useful to others. This feature is included in other plugins such as [python mode](https://github.com/python-mode/python-mode) so I think it would be a good addition.

### DIFF
--- a/plugin/rust.vim
+++ b/plugin/rust.vim
@@ -21,6 +21,16 @@ endif
 let &cpo = s:save_cpo
 unlet s:save_cpo
 
+if exists('g:cargo_run_key')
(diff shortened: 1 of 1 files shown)


da-x:
Thanks for you contribution!

First, if we are adding new configuration options, documentation for them is also needed. Second, I am not in favor of adding this because it does not really save on the amount of configuration that users need to do in order to add mappings. Users can already map commands with one-liners such as `nnoremap <Leader>r :Crun<CR>` in their config.

This reminds me, that we have not documented the `Cargo/Cbuild/Cargo` commands, and that deserves a separate PR!

Issue #196, opened 2018-02-23
syntastic broken 1.24


scirner22:
I recently upgraded to rust 1.24 and syntastic has been broken for me ever since. I use syntastic with the following config `let g:syntastic_rust_checkers = ['cargo']`.

```
NVIM v0.2.2
rustc 1.24.0
cargo 0.25.0
```

dmillard:
syntastic is also broken for me with the `cargo` checker. I'm a new user of rust.vim and thus don't have a point of comparison to old versions. The `rustc` checker does work, but gives extraneous errors on `extern crate ...` lines.

Interestingly for me, calling `:Errors` with the `cargo` checker shows up with the correct error list, implying syntastic has a good idea of what the errors _should_ be.

```rust
fn main() {
    garbage;
}
```

`:SyntasticInfo`
```
Syntastic version: 3.8.0-108 (Vim 800, Linux, GUI)
Info for filetype: rust
Global mode: active
Filetype rust is active
The current file will be checked automatically
Available checkers: cargo rustc
Currently enabled checker: cargo
```

`:let g:syntastic_debug=3 | SyntasticCheck`
```
syntastic: 13.622056: g:syntastic_version = '3.8.0-108 (Vim 800, Linux, GUI)'
syntastic: 13.622201: &shell = '/usr/bin/zsh', &shellcmdflag = '-c', &shellpipe = '2>&1| tee', &shellquote = '', &shellredir = '>%s 2>&1', &shelltemp = 1, &shellxquote = '', &autochdir = 0, &shellxescape = ''
syntastic: 13.622696: UpdateErrors: default checkers
syntastic: 13.622925: CacheErrors: default checkers
syntastic: 13.623481: g:syntastic_aggregate_errors = 0
syntastic: 13.623660: getcwd() = '<path>'
syntastic: 13.650012: system: command run in 0.024905s
syntastic: 13.650403: CacheErrors: Invoking checker: rust/cargo
syntastic: 13.651053: SyntasticMake: called with options: {'cwd': '<path>/src', 'errorformat': '%-G,%-Gerror: aborting %.%#,%-Gerror: Could not compile %.%#,%Eerror:%m,%Eerror[E%n]: %m,%Wwarning: %m,%Inote: %m,%C %#--> %f:%l:%c', 'makeprg': 'cargo check'}
syntastic: 13.953781: system: command run in 0.302281s
[...]

PR #204, opened 2018-04-12
set cwd path to project's root (containing Cargo.toml)

popravich:
This fixes issue #196 

Checked with:
```
rustc 1.25.0 (84203cac6 2018-03-25)
cargo 0.26.0 (41480f5cc 2018-02-26)
```
and syntastic master (vim-syntastic/syntastic@e3a819c7)

### DIFF
--- a/syntax_checkers/rust/cargo.vim
+++ b/syntax_checkers/rust/cargo.vim
@@ -41,7 +41,7 @@ function! SyntaxCheckers_rust_cargo_GetLocList() dict
 
     return SyntasticMake({
         \ 'makeprg': makeprg,
-        \ 'cwd': expand('%:p:h'),
+        \ 'cwd': fnamemodify(findfile("Cargo.toml", ".;"), ":p:h"),
         \ 'errorformat': errorformat })
 endfunction
 
--- a/syntax_checkers/rust/cargo.vim
+++ b/syntax_checkers/rust/cargo.vim
@@ -41,7 +41,7 @@ function! SyntaxCheckers_rust_cargo_GetLocList() dict
 
     return SyntasticMake({
         \ 'makeprg': makeprg,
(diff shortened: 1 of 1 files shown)


mitchmindtree:
I can confirm that this does also fixes the issue for me.

mitchmindtree:
Having a closer look, it looks like this depends on there being a Cargo.lock file at the root, which may not be the case for fresh projects. Perhaps it would be better to use Cargo.toml?

popravich:
I tried to use `Cargo.toml` in the first commit but it does not work for projects with workspaces.
It would be best if `cargo` had some command similar to `locate-project` returning path to root workspace.

tuzz:
This fixes the issue for me.

I'm setting `let g:syntastic_rust_checkers = ['cargo']` so it seems reasonable to check for the presence of `Cargo.lock`, but maybe this won't fix the issue for those using `rustc` directly.

I don't really understand why this broke in the first place?

popravich:
We can ensure that `Cargo.lock` is present by calling `cargo generate-lockfile`,
but my knowledge in programming Vim plugins is weak)

boblannon:
fixes the same issue for me

```
cargo 1.26.0 (0e7c5a931 2018-04-06)
rustc 1.26.2 (594fb253c 2018-06-01)
```

da-x:
Hi everyone, and thank you @popravich for your contribution!

I am not sure about relying on the existence of `Cargo.lock`. It is conventional in library crates not to include that file, and so I think it would be better to have Syntastic work correctly from the first build and not the second.

I think that the correct algorithm should be to look for a `Cargo.toml` upward, check if it belongs to a Cargo workspace, and if it doesn't, look again upward from there to find the `Cargo.toml` of the workspace. 

@popravich There are several issues opened on this problem, so please tell if you plan to make these changes :)

da-x:
Never mind, I've done the necessary changes. Thanks anyway!

PR #209, opened 2018-05-08
Fix error detection in rustfmt output

mjkoo:
Error behavior seems to have changed since
rust-lang-nursery/rustfmt@67285f15eb64637abbb134b675e04fd9a842d2f4

Fix regex to properly populate the location list


da-x:
Tested with a recent rustfmt. This indeed makes it work again, when there are errors, instead of echoing the content of the file (a separate issue :/ ). Thanks!

PR #215, opened 2018-06-10
Rustfmt not working on current nightly

korken89:
The issue was the use of an old flag compared to the new format: `--write-mode=overwrite` vs `--emit=files`
Fixed for current nightly, but I am not sure if this is the main error.

Please comment on the issue if it should be solved in another way.


korken89:
Saw there is a more general PR already, closing.

Issue #210, opened 2018-05-11
Syntastic 'cargo' cwd problems


MarkSwanson:
Syntastic is showing me a location list full of errors.
Unfortunately the pathnames for all of the errors are wrong. Cargo is presenting pathnames relative to the Cargo.lock file, but I've started vim like this:
$ cd cargo_member/src/bin
$ vi file.rs

So the location list contains:
cargo_member/src/bin/file.rs

But my CWD in vim is already cargo_member/src/bin

So when I ":lnext" syntastic/vim fails to find the file because it's looking for: cargo_member/src/bin/cargo_member/src/bin/file.rs

Solution?: 'fix' the file paths that Cargo returns to be relative to Vim's CWD.
  We know Cargo will include everything up to the dir containing Cargo.lock. Maybe it would be easy to remove the set of dirs in the pathname based on the CWD. 
  example: change %f in the errorformat from cargo_member/src/bin/file.rs to be just file.rs (if the CWD contains cargo_member/src/bin/)

doomrobo:
Hey, would you be able to check if #218 fixes your issue? Thanks!

da-x:
I've done a related fix in master - Syntastic should work regardless of the current directory. Would you please test and tell if it works for you? Thanks!

PR #218, opened 2018-06-17
Make vim quickfix aware of root directory of crate

doomrobo:
Previously, the quickfix window would be unaware of the current working directory. So if my cwd was `CRATEROOT/src` and quickfix parsed cargo's output as having an error at `src/mem.rs`, then I would be taken to `CRATEROOT/src/src/mem.rs`, which doesn't exist.

The changes I made make quickfix aware of the current directory so that error following now works in most cases. I haven't tested it, but this may be a fix for #196 and/or #210.

Notably, this is limited. The plugin will treat `Compiling/Checking CRATE v0.0.0 (file://PATH)` as a trigger that `PATH` is the cwd. But if several things are being compiled/checked at the same time, quickfix will just assume the most recent message is the cwd. This is not always correct. However, recompiling will often reduce the number of modules in the output to about 1. You can test out this functionality on any crate with local submodules (I used alacritty).

### DIFF
--- a/compiler/rustc.vim
+++ b/compiler/rustc.vim
@@ -40,6 +40,8 @@ CompilerSet errorformat+=
 			\%Eerror[E%n]:\ %m,
 			\%Wwarning:\ %m,
 			\%Inote:\ %m,
+			\%D%\\s%#Checking\ %.%#(file://%f),
+			\%D%\\s%#Compiling\ %.%#(file://%f),
 			\%C\ %#-->\ %f:%l:%c
 
 let &cpo = s:cpo_save


da-x:
Hi, and thanks for your contribution!

This change would not be valid in case of Cargo workspaces, as in that particular case the printed source paths are relative to the root of the workspace and not the root of each crate. I think the approach of having the current directory match the output of what you are executing is a better fix. 

Externally, some take care of that by using plugins that set the current directory to the root of the Git project. However instead of relying on additional plugins perhaps we can still have something working internally that improves things.

If you are using Syntastic to build then perhaps an improved version of the fix in #204 will be good for all cases. Otherwise if you are just using `:make` then we can probably come up with a shortcut to facilitate this. What do you think?

da-x:
I've done some related fix in `master`. This makes the PR unnecessary - but thanks anyway. Would you please test the master branch and tell if it works?

doomrobo:
> This change would not be valid in case of Cargo workspaces, as in that particular case the printed source paths are relative to the root of the workspace and not the root of each crate.

Nice catch! I made a small test case to see your point. I made a crate called `wkspacetest` and a workspace called `submod` that fails to compile. I agree that the plugin doesn't work (I just tried), but it fails for the opposite reason: the error messages look like this
```
Compiling submod v0.1.0 (file:///private/tmp/wkspacetest/submod)
error[E0425]: cannot find function `blah` in this scope
--> submod/src/lib.rs:2:5
    |
  2 |     blah()
    |     ^^^^ not found in this scope

error: aborting due to previous error
```
It fails because the plugin detects the `Compiling ... (file://` line and sets that as the cwd, but the file path `submod/src/lib.rs` is relative to the *root* crate as opposed to the workspace.

Anyways, I don't use Syntastic at all, so unfortunately #204 doesn't help me. I'm sure something like what you did already (looking for the nearest `Cargo.toml` that isn't part of a workspace) could work here, but I'd have to take a closer look at quickfix docs.

da-x:
Are you using `:make`?

PR #213, opened 2018-05-19
Add an option to let rustfmt 0.7 work

Marwes:
Added an option to let this plugin work with both rustfmt < 0.6 and >=
0.7.

Fixes #212

PR #221, opened 2018-06-25
Use g:rustfmt_command for the version check

Marwes:
As pointed out by @jonhoo in https://github.com/rust-lang/rust.vim/pull/213#discussion_r197833588 . We need to use the user specified command for all rustfmt invocations.


da-x:
Thanks!

PR #222, opened 2018-06-26
Add support for the dyn keyword highlighting

### LINKED ISSUE #219: The contextual "dyn" keyword is not recognized
With the new `dyn Trait` syntax in Rust 1.27, rust.vim doesn't recognize `dyn` as a keyword.

Since it's contextual, I looked up where [`dyn` is parsed](https://github.com/rust-lang/rust/blob/3eda71b00ad48d7bf4eef4c443e7f611fd061418/src/libsyntax/parse/parser.rs#L1585-L1592):

```rust
        } else if self.check_keyword(keywords::Dyn) &&
                  self.look_ahead(1, |t| t.can_begin_bound() &&
                                         !can_continue_type_after_non_fn_ident(t)) {
            self.bump(); // `dyn`
            // Always parse bounds greedily for better error recovery.
            let bounds = self.parse_ty_param_bounds()?;
            impl_dyn_multi = bounds.len() > 1 || self.prev_token_kind == PrevTokenKind::Plus;
            TyKind::TraitObject(bounds, TraitObjectSyntax::Dyn)
```

And here's the source for [`can_begin_bound`](https://github.com/rust-lang/rust/blob/3eda71b00ad48d7bf4eef4c443e7f611fd061418/src/libsyntax/parse/token.rs#L313-L317) and [`can_continue_type_after_non_fn_ident`](https://github.com/rust-lang/rust/blob/3eda71b00ad48d7bf4eef4c443e7f611fd061418/src/libsyntax/parse/parser.rs#L403-L411).

da-x:
Fixes #219

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -153,6 +153,16 @@ syn region    rustDerive      start="derive(" end=")" contained contains=rustDer
 " Some are deprecated (Encodable, Decodable) or to be removed after a new snapshot (Show).
 syn keyword   rustDeriveTrait contained Clone Hash RustcEncodable RustcDecodable Encodable Decodable PartialEq Eq PartialOrd Ord Rand Show Debug Default FromPrimitive Send Sync Copy
 
+" dyn keyword: It's only a keyword when used inside a type expression, so
+" we make effort here to highlight it only when Rust identifiers follow it 
+" (not minding the case of pre-2018 Rust where a path starting with :: can
+" follow).
+"
+" This is so that uses of dyn variable names such as in 'let &dyn = &2' 
+" and 'let dyn = 2' will not get highlighted as a keyword.
+syn match     rustKeyword "\<dyn\ze\_s\+\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)" contains=rustDynKeyword
+syn keyword   rustDynKeyword  dyn contained
+
 " Number literals
 syn match     rustDecNumber   display "\<[0-9][0-9_]*\%([iu]\%(size\|8\|16\|32\|64\|128\)\)\="
 syn match     rustHexNumber   display "\<0x[a-fA-F0-9_]\+\%([iu]\%(size\|8\|16\|32\|64\|128\)\)\="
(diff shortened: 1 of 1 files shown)


da-x:
@chris-morgan I think you're an expert on Vim syntax highlighting - did I get it right?

syntax/rust.vim
```diff
@@ -153,6 +153,14 @@ syn region    rustDerive      start="derive(" end=")" contained contains=rustDer
 " Some are deprecated (Encodable, Decodable) or to be removed after a new snapshot (Show).
 syn keyword   rustDeriveTrait contained Clone Hash RustcEncodable RustcDecodable Encodable Decodable PartialEq Eq PartialOrd Ord Rand Show Debug Default FromPrimitive Send Sync Copy
 
+" dyn keyword: It's only a keyword when used inside a type expression,
+" with two options: detected by a preceding '<' or, or a preceding '&'
+" as long the next token is an identifier. This is so that uses of dyn 
+" variable names such as in 'let &dyn = &2' and 'let dyn = 2' will not 
+" get highlighted as a keyword.
+syn match     rustKeyword "[&<] *\(dyn\) \+[a-zA-Z_]"me=e-1,he=e-1 contains=rustOperator,rustDynKeyword,rustSigil
```
chris-morgan: A type can have multiple generic parameters, e.g. `Foo<dyn Bar, dyn Baz>`. Adding `,` to `[&<]` would resolve that.

But then there are other things that can precede it as well, e.g. `mut` (`&mut dyn Foo`) or a lifetime (`&'foo dyn Foo`).

In normal syntax it’s an error to have a path following an identifier, so I think just dropping the preceding token part would be sufficient. I think we can ignore paths with leading double colon (`dyn ::std::fmt::Display`), since that’s on the way out in Rust 2018 anyway. Weirdly-placed comments are always going to mess up a simple highlighter like this, so let’s not try to go crazy about that.

I’d prefer to use the start of rustIdentifier rather than `[A-Za-z_]`; that way we keep support for non-ASCII identifiers.

All up, I suggest `syn match rustKeyword "\<dyn\ze\_s\+\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)" contains=rustDynKeyword` (and a tweaking of the comment to match, of course).

syntax/rust.vim
```diff
@@ -244,6 +252,7 @@ hi def link rustFloat         Float
 hi def link rustArrowCharacter rustOperator
 hi def link rustOperator      Operator
 hi def link rustKeyword       Keyword
+hi def link rustDynKeyword    Keyword
```
chris-morgan: Link it to `rustKeyword` rather than `Keyword`.

PR #224, opened 2018-06-29
autodetect Cargo.toml filetype as cfg

hrls:
Some highlighting at least

### DIFF
--- a/ftdetect/rust.vim
+++ b/ftdetect/rust.vim
@@ -1,1 +1,2 @@
 au BufRead,BufNewFile *.rs set filetype=rust
+au BufRead,BufNewFile Cargo.toml set filetype=cfg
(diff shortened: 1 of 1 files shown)


da-x:
Any reason why `cfg` is better than using [vim-toml](https://github.com/cespare/vim-toml)? I think we should at least make this optional, right? So it would not conflict with `vim-toml`.

hrls:
@da-x `cfg` isn't better than `vim-toml`. Its just for some kind of highlighting without any additional vim plugins.

chris-morgan:
I have a suggestion:

```vim
au BufRead,BufNewFile Cargo.toml if &ft == "" | set filetype=cfg | endif
```

That way, if something else has already set a filetype (e.g. if vim-toml is earlier in the runtimepath), it will leave it alone, but if nothing has (e.g. if vim-toml is later in the runtimepath or missing) it will set it to cfg; vim-toml will then just clobber that with toml, and everyone will be happy.

da-x:
@chris-morgan - very good suggestion :) I tested just now with `vim-toml` on/off before/after, seems to work. @hrls can you please make this fix?

hrls:
@chris-morgan Thanks!
@da-x Sure, done.

da-x:
Thanks!

PR #227, opened 2018-07-01
Support crate keyword.

davidtwco:
In the 2018 edition, the `crate` keyword [is valid as a visibility modifier](https://rust-lang-nursery.github.io/edition-guide/2018/transitioning/modules/path-clarity.html). This PR modifies the syntax highlighting to reflect this (mostly because `rustc` uses it and seeing it mishighlighted has been bugging me).

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -27,6 +27,7 @@ syn match     rustPanic       "\<panic\(\w\)*!" contained
 syn keyword   rustKeyword     break
 syn keyword   rustKeyword     box nextgroup=rustBoxPlacement skipwhite skipempty
 syn keyword   rustKeyword     continue
+syn keyword   rustKeyword     crate
 syn keyword   rustKeyword     extern nextgroup=rustExternCrate,rustObsoleteExternMod skipwhite skipempty
 syn keyword   rustKeyword     fn nextgroup=rustFuncName skipwhite skipempty
 syn keyword   rustKeyword     in impl let
@@ -42,8 +43,6 @@ syn keyword   rustKeyword     use nextgroup=rustModPath skipwhite skipempty
 syn keyword   rustKeyword     mod trait nextgroup=rustIdentifier skipwhite skipempty
 syn keyword   rustStorage     move mut ref static const
 syn match rustDefault /\<default\ze\_s\+\(impl\|fn\|type\|const\)\>/
-
-syn keyword   rustInvalidBareKeyword crate
 
 syn keyword rustPubScopeCrate crate contained
 syn match rustPubScopeDelim /[()]/ contained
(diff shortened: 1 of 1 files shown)


syntax/rust.vim
```diff
@@ -32,6 +32,7 @@ syn keyword   rustKeyword     fn nextgroup=rustFuncName skipwhite skipempty
 syn keyword   rustKeyword     in impl let
 syn keyword   rustKeyword     macro
 syn keyword   rustKeyword     pub nextgroup=rustPubScope skipwhite skipempty
+syn keyword   rustKeyword     crate skipwhite skipempty
```
chris-morgan: `skipwhite skipempty` is only needed (or desirable) if `nextgroup` is specified. Also note that all these keywords are in alphabetical order so far, so putting this one with or after `continue` (I have no particular preference) would be good.

syntax/rust.vim
```diff
@@ -43,8 +44,6 @@ syn keyword   rustKeyword     mod trait nextgroup=rustIdentifier skipwhite skipe
 syn keyword   rustStorage     move mut ref static const
 syn match rustDefault /\<default\ze\_s\+\(impl\|fn\|type\|const\)\>/
 
-syn keyword   rustInvalidBareKeyword crate
-
```
chris-morgan: Remember to remove its `hi def link` line too.

syntax/rust.vim
```diff
@@ -154,11 +153,11 @@ syn region    rustDerive      start="derive(" end=")" contained contains=rustDer
 syn keyword   rustDeriveTrait contained Clone Hash RustcEncodable RustcDecodable Encodable Decodable PartialEq Eq PartialOrd Ord Rand Show Debug Default FromPrimitive Send Sync Copy
 
 " dyn keyword: It's only a keyword when used inside a type expression, so
-" we make effort here to highlight it only when Rust identifiers follow it 
+" we make effort here to highlight it only when Rust identifiers follow it
```
chris-morgan: This and the one below it are not pertinent to the matter. Ideally, all the crate stuff will be in one commit, and this other stuff will either not be done, or be in a separate commit. Practice good commit hygiene! It’s fun! Sometimes it’s even useful afterwards! ☺

davidtwco: Oops, I didn't even notice this, the editor did it automatically. I'll separate it out.

syntax/rust.vim
davidtwco: Should be fixed now.

syntax/rust.vim
davidtwco: Should be fixed now.

syntax/rust.vim
chris-morgan: A few years ago I got tired of that sort of thing happening, and so I started always using `--patch` on my `git add` invocations when modifying existing files (when adding new files you can’t), and verifying that the diff is in fact what I want to commit. With aliases, this means I now type `g ap`.

The staging area can definitely be helpful if you deliberately make it a part of your process.

da-x: Sorry, it was my editor that introduced those extraneous white spaces. I recently moved to vim-plug and forgot to re-enable `bronson/vim-trailing-whitespace`.

PR #233, opened 2018-07-08
Indentation fixes plus tests

da-x:
@chris-morgan this fixes half of the indentation issues (the nagging ones).

I also run a full indent test on a large file from some project, besides the unit tests, and diffed the results. Going to do more tests to make sure.

### DIFF
--- a/indent/rust.vim
+++ b/indent/rust.vim
@@ -81,8 +81,17 @@ function! s:is_string_comment(lnum, col)
     endif
 endfunction
 
-function GetRustIndent(lnum)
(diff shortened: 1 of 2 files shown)


da-x:
This has been long overdue.

The major difference is that 'where' should now indent properly. 

It's not perfect, but it should bring us much closer to rustfmt. Observe for example, that it full indentation via Vim causes much less ruin to large example files such as https://github.com/bodil/im-rs/blob/master/src/ord/map.rs .

PR #240, opened 2018-07-11
indent: fix a cindent failure

BurntSushi:
This commit attempts to fix what appears to be a failure mode in
cindent. Here's a minimal (but silly) example:

```
fn main() {
    if true {
{}
    }
}
```

In the above code, inserting a new line after the penultimate brace
dedents out to the first column instead of retaining the indent of the
previous line. The dedent to the first column persists regardless of the
nesting depth, e.g.,

```
fn main() {
    if true {
        if true {
{}
        }
    }
}
```

The behavior here is the same: inserting a new line after the
penultimate brace dedents the cursor back to the first column.

This example seems contrived, but the way I noticed it was in code that
roughly looked like this:

```
fn main() {
    if true {
        println!("
    Something happened:
{}
    ", "wat");
    }
}
```

in which the same failure mode persists.

This commit attempts to fix the issue by short-circuiting and using the
previous line's indent level when following two conditions are met:

    1. The last character of the previous line is a closing brace.
    2. cindent would otherwise dedent to the first column.

In this case, I claim that we always want to retain the indent of the
previous line (the closing brace). In the case where the correct dedent
is to the first column, then we retain that behavior because it should
imply that the indent of the previous line is itself the first column.

N.B. This is my first foray with investigating and attempting to fix an
indent bug in Vim. Please scrutinize carefully. :-)

### DIFF
--- a/indent/rust.vim
+++ b/indent/rust.vim
@@ -133,6 +133,9 @@ function GetRustIndent(lnum)
                 \ && prevline =~# '^\s*where\s'
         return indent(prevlinenum) + 6
     endif
+    if prevline[len(prevline) - 1] ==# "}" && cindent(a:lnum) == 0
+        return indent(prevlinenum)
+    endif
 
     if prevline[len(prevline) - 1] ==# ","
                 \ && s:get_line_trimmed(a:lnum) !~# '^\s*[\[\]{}]'


BurntSushi:
Oy. This PR does not fix my bug. The dedent happens regardless of what the previous line ends with, so checking for just the closing brace is too conservative.

da-x:
Hey @BurntSushi, would you like to test the [rewritten-indent](https://github.com/rust-lang/rust.vim/tree/rewritten-indent) branch for awhile? I've rewritten the indent function in Python, as it's available on many Vim setups. The new implementation is independent from `cindent` and fixes all issues including this one.

BurntSushi:
@da-x \o/ my hero! I will try that next time tonight or tomorrow and report back. :-)

Is there a path for that branch getting on to master?

da-x:
It can go in master after I'm convinced that enough Vim setups have Python support (2 or 3, does not matter), and that the implementation is not heavy on resources (memory or CPU). At any case, the changes keep fallback to the old implementation in case Vim does not support Python.

Originally I sought a vimscript implementation, but figuring out token trees was too heavy on it.

da-x:
@phansch, are you having good experience with the `rewritten-indent` branch?

phansch:
@da-x Yes, absolutely. The only thing that I noticed is the following:

https://asciinema.org/a/3FYoMwY3vnMECWdFVy4I58Qap

The closing `)` of function/macro calls doesn't seem to be indented correctly automatically, but I'm not sure if that's in the scope of this issue.

da-x:
Thanks. At least it's not a regression - seems to happen without the branch too.

PR #246, opened 2018-07-29
fix: Prevent the rustfmt autocommand from triggering itself

### LINKED ISSUE #245: Opening multiple buffers causes rustfmt on save to become slow
* rust.vim version: a247c383ad35de70e1d0481041b2c63dec5d47da

Steps to reproduce: Save a file with rustfmt on save enabled.

Expected vs. actual behavior: (neo)vim seems to try and write to disk multiple times judging from the disk activity. I'd expect only a single write to happen. Opening multiple files seems to increase the number of writes done so perhaps all currently opened buffers are formatted and written to disk?

This happens in both vim and neovim (but neovim becomes slower quicker)

```
rust.vim Global Variables:

let g:rust_conceal = v:null
let g:rust_fold = v:null
let g:rust_last_args = v:null
let b:rust_last_args = []
let g:rustc_path = v:null
let g:rustfmt_autosave = 1
let g:rustfmt_emit_files = 1
let g:rustfmt_options = ''
rustfmt 0.8.2-nightly (5e59925 2018-07-02)

rustc 1.29.0-nightly (12ed235ad 2018-07-18)

cargo 1.29.0-nightly (af9e40c26 2018-07-05)


NVIM v0.2.2
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: /nix/store/4r5kszyy0iirc5agfah45lvz7mnnsrb4-gcc-wrapper-7.3.0/bin/gcc -Wconversion -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -O2 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -I/build/source/build/config -I/build/source/src -I/nix/store/76rnm9q9y5ja4pxj84h0pyim320vlihr-libuv-1.19.2/include -I/nix/store/h4s5w59nsnvc3d5xs5jjmmjyvzkz0n2s-libmsgpack-2.1.5/include -I/nix/store/c9jmn5dd955zigj9czfwlp58a4fm43cc-unibilium-2.0.0/include -I/nix/store/bxa8yy7xb4b44a85knplybibxbjvqb3h-libtermkey-0.20/include -I/nix/store/x140lpahjmp9i54isa3mfc57j2dmbrj3-neovim-libvterm-2017-11-05/include -I/nix/store/nr3har95xnnr2ip8yx01s7gbj753njvx-jemalloc-5.0.1/include -I/nix/store/fn2x7v3bs305xl55ynj698mjc0yfn1lh-glibc-2.26-131-dev/include -I/build/source/build/src/nvim/auto -I/build/source/build/include
Compiled by nixbld

Features: +acl +iconv +jemalloc +tui 
See ":help feature-compile"

  fall-back for $VIM: "/nix/store/8q7phn3alha1jkf7lazad4pwdjg4jgaj-neovim-unwrapped-0.2.2/share/nvim"

Run :checkhealth for more info
```

Marwes:
Introduced in 2fb3df5, the writefile call would trigger the rustfmt
autocommand again if multiple Rust files were opened. setline prevents
this from happening and the `silent edit!` seems to be enough to remedy
the jump-to-start-of-file issue that 2fb3df5 fixed.

Fixes #245

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -146,7 +146,7 @@ function! s:RunRustfmt(command, tmpname, fail_silently)
             let l:content = readfile(a:tmpname)
         endif
 
-        call writefile(l:content, expand('%'))
+        call setline(1, l:content)
         silent edit!
         let &syntax = &syntax
 


da-x:
Wow thanks, I didn't know `writefile()` has this side effect... thought it was a 'pure' function that did not care about open buffer states. Plus, the Vim doc does not mention this behavior.

As for `setline(1, ...)`, if the result of the formatting is shorter in the number of lines currently in the buffer, then there would be a leftover from the old content between `len(content)` and the current number of lines, which would need to be truncated off. There no function that can be used, other than a scripted invocation of `:delete`, for which we need to make sure there are no side-effects.

da-x:
That's actually what we had before 2fb3df5c537, but it negatively affected the current line and the undo history.

Marwes:
I rechecked it again and it seems that the `1,$d _` is what caused the view to jump to line 1. `silent edit!` inserts an extra place one needs to undo past (and empty edit) which seems fine.

The issue of jumping to line 1 does not exist anyway, no idea why `1,$d _` were there in the first place.

Marwes:
> As for setline(1, ...), if the result of the formatting is shorter in the number of lines currently in the buffer, then there would be a leftover from the old content between len(content) and the current number of lines, which would need to be truncated off. There no function that can be used, other than a scripted invocation of :delete, for which we need to make sure there are no side-effects.

This does seem to happen at least :man_shrugging: , the file gets truncated as it should be.

Marwes:
DONT MERGE, I seem to have broken something actually.

da-x:
No worries :) it's hard as it is to get this functionality implemented right.

I would suspect that with the PR as it is that `silent edit!` reverts the new reformatted content of the buffer with the old content from the file. Is that what you're getting?

Marwes:
Yep, removing the silent edit fixes on thing, but you are also correct in that the file needs truncating as that does not seem to happen (except sometimes I guess as it seemed to work before).

Marwes:
Haven't been able to figure out a solution which preserves the location, does not cause multiple saves, truncates the file correctly and preserves the file modes :/ At least one of those issues are always present.

da-x:
Does master fix it now for you?

Marwes:
@da-x Yup, that fixes it. Seems I was chasing the wrong problem.

PR #257, opened 2018-09-03
b:rust_compiler_cargo_qf_prev_cd_saved not defined issue

dbger:
It's show b:rust_compiler_cargo_qf_prev_cd_saved not defined message when add blow content in vimrc
```
" open quickfix window at onece
autocmd QuickFixCmdPost [^l]* nested cwindow
autocmd QuickFixCmdPost    l* nested lwindow
```


da-x:
Thanks. It would be good to know in what cases `QuickFixCmdPost` can be called on a buffer without a preceding `QuickFixCmdPost` with the same buffer.

Issue #259, opened 2018-09-07
Neomake - workspaces are broken


dpc:
I'm opening this issue just to get some visibility of people that are both Rust user and VimL-versed.

https://github.com/neomake/neomake/issues/1916

Let's say you have project `a` with `a/b` subcrate.

If you `cd a/b` and `vim`, neomake stops working well, because `rustc` still displays error paths as `b/src/lib.rs` (relative to workspace root) while neomake/Vim would like them to be relative to current dir, so just `src/lib/rs`.

All it takes to fix (I think) is fixing the path somewhere here: https://github.com/neomake/neomake/blob/master/autoload/neomake/makers/ft/rust.vim#L136 by stripping the prefix.

But it's currently beyond me, how to do it right, and I don't have time to dedicate to figure it out.

Please, someone take a look. :)

dpc:
The same problem happens eg. in VSCode BTW.

blueyed:
As for Neomake, it should probably `cd` to the project root?
Currently it `cd`s to the parent dir of the buffer (https://github.com/neomake/neomake/blob/28f6991f3546195e764052d5e1c731432ac8f706/autoload/neomake/makers/ft/rust.vim#L28), which also means that it should not make a difference if you `cd` yourself (this is for the cargo maker, but not rustc for example!). Please consider filing an issue in Neomake as a followup if e.g. this is not working with `rustc` for you, but with cargo.

dpc:
@blueyed I'd like to help, but I am so confused that I don't know how to answer any part of your question.

> As for Neomake, it should probably cd to the project root?

The problem is that with Rust workspaces the is "root workspace dir" and bunch of "root crate dirs". inside it. When I use `rust.vim` + `neomake` in a typical project that has one crate (and thus workspace root == crate root) I can `cd` anywhere in the project, and `:ll` jumps to error just fine.

When I start a workspace things work only if the current dir is the workspace root.

Not sure if it's any help.

blueyed:
Does it also affect `cargo`, or only `rustc`?  (the former does `cd` in Neomake, the latter does not)
You can check `:echom getloclist(0)` to see the parsed location list.
It looks like an issue in Neomake's makers, which need to be adjusted probably, but we need to figure out which ones are affected.
I think cd'ing to `%:p:h` is not the best probably (i.e. it might not match the maker's output).
[...]

Issue #236, opened 2018-07-09
Rustfmt on save causes undo to jump to beginning of file


Marwes:
* rust.vim version: https://github.com/rust-lang/rust.vim/commit/3f5bb1d2b1ec7d26b7a9e871eaf9738230630442

### Steps to reproduce:

Make a change in a file then save so rustfmt (on save) is run. Then undo the change.

### Expected vs. actual behavior:

Undoing the change(s) should make the view stay at where the manual change was made. What happens instead is that the view jumps to the beginning of the file.

Bisected to one of the following commits (formatting was broken in many of them so I don't know which exactly).

a5290d6ec2e8da8474e8eaf3aa3b2847b7f66e97
f1d785a3135d3187f22d79823e0fa995cbcffdbc
ba190eecc64a6c7103a569f23fb223f69e49869d
ff54b40bbcb2034f98283fbe4bbccc0e9d57f6f4
6b4eb5d9ceb0b2437966adfcb1fd2064403cebc3
c7c99f3d8fcdbda1be03a7ebee109609605f681f
aff86a1a11245bf7ad641c40a08b01922b59dc1f

```
rust.vim Global Variables:

let g:ftplugin_rust_source_path = v:null
let g:loaded_syntastic_rust_cargo_checker = v:null
let g:loaded_syntastic_rust_filetype = v:null
let g:loaded_syntastic_rust_rustc_checker = v:null
let g:rust_bang_comment_leader = v:null
let g:rust_cargo_avoid_whole_workspace = v:null
let g:rust_clip_command = v:null
let g:rust_conceal = v:null
let g:rust_conceal_mod_path = v:null
let g:rust_conceal_pub = v:null
let g:rust_fold = v:null
let g:rust_last_args = v:null
let b:rust_last_args = []
let g:rust_last_rustc_args = v:null
let b:rust_last_rustc_args = []
let g:rust_original_delimitMate_excluded_regions = v:null
let g:rust_playpen_url = v:null
let g:rust_recent_nearest_cargo_tol = v:null
let g:rust_recent_root_cargo_toml = v:null
let g:rust_recommended_style = v:null
let g:rust_set_conceallevel = v:null
let g:rust_set_conceallevel=1 = v:null
let g:rust_set_foldmethod = v:null
let g:rust_set_foldmethod=1 = v:null
let g:rust_shortener_url = v:null
let g:rustc_makeprg_no_percent = v:null
let g:rustc_path = v:null
[...]

PR #261, opened 2018-09-09
Fix `cwd` issues when using Neomake

### LINKED ISSUE #260: Automatic rustfmt erases my files sometimes.
* rust.vim version: Current one, I use vim-plug and do `:PlugUpdate` often.

Steps to reproduce:

I think it started happening when I converted my project to use a workspace. The current state is: https://github.com/dpc/crev/commit/d9fbf06bb1385181b3788fc0cffd544ec66ff565

Expected vs. actual behavior:

Typically my vim consistently reformats a file on every save.

Now, for whatever reason, it stops working in the middle of the session for some files (for which it worked fine just a second ago) and from that time it just keeps erasing them every time I write the buffer. :/

Paste debugging info from the Rust Vim plugin via _one_ of the following
commands: `:RustInfo`,  `:RustInfoToClipboard`, or `:RustInfoToFile <filename>`.

```
rust.vim Global Variables:

let g:rust_conceal = v:null
let g:rust_fold = v:null
let g:rust_last_args = v:null
let b:rust_last_args = []
let g:rustc_path = v:null
let g:rustfmt_autosave = 0
let g:rustfmt_autosave_if_config_present = 1
let g:rustfmt_emit_files = 1
let g:rustfmt_options = ''
rustfmt 0.99.4-nightly (1c40881 2018-08-27)

rustc 1.30.0-nightly (fc81e3624 2018-09-07)

cargo 1.29.0-nightly (0ec7281b9 2018-08-20)


NVIM v0.3.1
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: /nix/store/wkfafgwgs1na5f808wybs8l6v8yjrvqz-gcc-wrapper-7.3.0/bin/gcc -Wconversion -O2 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -Wno-array-bounds -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/build/source/build/config -I/build/source/src -I/nix/store/3jzza70h4xlcl65py98y0ddq3sr3nv6z-libuv-1.21.0/include -I/nix/store/wqmabb91mn7c8l0cbagbi7n6b78r2k6g-libmsgpack-2.1.5/include -I/nix/store/qwlgay0pdgllmdwif144771z8y0anx3j-unibilium-2.0.0/include -I/nix/store/caygb8bjk106ja8fddlx2zw5fbzzz8g2-libtermkey-0.20/include -I/nix/store/r54qm8js4ndph7mqwm61fgpd54niwbxi-neovim-libvterm-2017-11-05/include -I/nix/store/k8fcymm135rp923s2w8w5z7wzr2kdcvi-jemalloc-5.0.1/include -I/nix/store/z10kkr081zb02kyjw8ybf2s9zll49cvx-glibc-2.27-dev/include -I/build/source/build/src/nvim/auto -I/build/source/build/include
Compiled by nixbld

Features: +acl +iconv +jemalloc +tui 
See ":help feature-compile"

  fall-back for $VIM: "/nix/store/19sgqgn1vpc8c7828hhf0lmpaqx9yb8m-neovim-unwrapped-0.3.1/share/nvim"

Run :checkhealth for more info
```

dpc:
All credit to @blueyed .

This fixes #260

The investigation is in the corresponding Neomake issue:
https://github.com/neomake/neomake/issues/2080

This is also related to: #259

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -146,8 +146,8 @@ function! s:RunRustfmt(command, tmpname, fail_silently)
             let l:content = readfile(a:tmpname)
         endif
 
-        call writefile(l:content, expand('%'))
-        silent edit!
+        normal! "_ggdG
+        call setline(1, l:content)
         let &syntax = &syntax
 
         " only clear location list if it was previously filled to prevent


autoload/rustfmt.vim
```diff
@@ -146,8 +146,8 @@ function! s:RunRustfmt(command, tmpname, fail_silently)
             let l:content = readfile(a:tmpname)
         endif
 
-        call writefile(l:content, expand('%'))
-        silent edit!
+        normal! ggdG
+        call setline(1, l:content)
         let &syntax = &syntax
```
blueyed: The `let &syntax = &syntax` is also not necessary anymore likely (if it ever was, but maybe due to the `edit` that was used previously).

TLDR of the investigation: the problem is that `expand('%')` here might be relative, but it was `:lcd`ed before.
An alternative might be to use the full buffer name (save it before `cd`ing), but I think just setting the lines is better than writing from outside.

blueyed:
btw: as can be seen in https://github.com/neomake/neomake/issues/2080#issuecomment-419692234 the error/issue shows up immediately when not using `:silent!` - therefore this should really be avoided, since it hides all kind of errors.

autoload/rustfmt.vim
```diff
@@ -146,8 +146,8 @@ function! s:RunRustfmt(command, tmpname, fail_silently)
             let l:content = readfile(a:tmpname)
         endif
 
-        call writefile(l:content, expand('%'))
-        silent edit!
+        normal! ggdG
```
blueyed: Just noticed that this should become `normal! "_ggdG` to not mess with registers.

autoload/rustfmt.vim
blueyed: Actually `expand('%')` should be the relative path after `:lcd` - so there is some other reason for this (maybe a bug in Vim/Neovim even (NVIM v0.3.1 used by the reporter)).

Anyway, the `cd`ing back should be done in a `try … finally` block, because otherwise it might not change it back.

blueyed:
Please remove the mentioning of me before merging it, otherwise it will show up in notifications all over the place.  Use `@blueyed` maybe in backticks or so.

blueyed:
Thanks for trying - I was still notified of the new commits though.. :/

da-x:
Hi @blueyed and @dpc ,

Thanks for debugging the issue.

However the problem with the `normal! "_ggdG` and `call setline(1, l:content)` sequence is that it's a regression over issue #236. We should also prevent undo actions from being affected by this fix.

blueyed:
@da-x 
I see (https://github.com/rust-lang/rust.vim/commit/2fb3df5c53770d7710d0171a6d5ea7767f86bce9).
I've looked at neoformat, and they use a similar approach: https://github.com/sbdchd/neoformat/blob/master/autoload/neoformat.vim#L170
Maybe it is more related to the (missing/wrong) undojoin, and or `mkview!` etc?

blueyed:
@dpc 
Not sure.  I think it still is a bit fragile what is being done in rust.vim, but there appear to be reasons, and I am not using it myself anyway.

I suggest closing this for now - it can get re-opened later.

da-x:
@dpc @blueyed Looking over neoformat, I pushed 4c21791115d44e to master. It should do something similar this PR, but without regressing over #236. The existing `undojoin` seems to work with this change without any cursor funkiness on `u`. I wished though that `winsaveview` and `winrestview` would work right like in neoformat but was not able to achieve that. Feel free to test.

blueyed:
@da-x 
Looks good, thanks!

dpc:
I've updated all vim plugins to latest and will keep working on my workspace-enabled project. I'll report any issues. Thanks everyone for looking into this!

PR #265, opened 2018-09-16
Add support for 'cargo check --all-features'

vtavernier:
This PR adds support for the `--all-features` option of `cargo check`, which is required if the current project has code which is conditionally compiled depending on features, but we would still like to get syntax checking for said code.


da-x:
Thanks!

PR #267, opened 2018-09-20
Add :RustTest to run a test under the cursor

rhysd:
I added a new command `:RustTest`.

From doc:

```
:RustTest[!] [options]                                          *:RustTest*
		Runs a test under the cursor when the current buffer is in a
		cargo project with "cargo test" command. If the command did
		not find any test function under the cursor, it stops with an
		error message.

 		When ! is given, runs all tests regardless of current cursor
		position.

 		When [options] is given, it is passed to "cargo" command
		arguments.

 		When the current buffer is outside cargo project, the command
		runs "rustc --test" command instead of "cargo test" as
		fallback. All tests are run regardless of adding ! since there
		is no way to run specific test function with rustc. [options]
		is passed to "rustc" command arguments in the case.
```

The motivation of this command is to run a test quickly and make test/fix loop faster. By running only test under the cursor, a developer can confirm the result of the specific test just after fixing it. This command is especially effective when a project is bigger and running all tests take longer time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,11 @@ If you set g:rust_clip_command RustPlay will copy the url to the clipboard.
 
         let g:rust_clip_command = 'xclip -selection clipboard'
 
+### Running a test under cursor
+
+In cargo project, the `:RustTest` command will run a test under the cursor.
+This is useful when your project is bigger and running all tests take longer time.
+
 ## Help
 
 Further help can be found in the documentation with `:Helptags` then `:help rust`.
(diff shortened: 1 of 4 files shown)


rhysd:
Hello, how about this PR? I've been using this branch and it works fine. If you're super busy and you don't have take much time to maintain, it's ok for me to help maintenance of this repository.

da-x:
Sorry @rhysd, I've been out vacationing for awhile. BTW if you want to join as maintainer you are more than welcomed! Maybe msg @steveklabnik about it :)

rhysd:
No problem! Thank you for checking this PR soon. I'm sorry for rushing you.

> BTW if you want to join as maintainer you are more than welcomed! Maybe msg @steveklabnik about it :)

I understood. I might not be so active, but I think I can help this project. I'll consider. Thank you for your advice.

PR #268, opened 2018-09-21
Create new syntax name for control flow keywords

mandreyel:
I'm writing a vim color scheme and I'd like to be able to differentiate `return`, `break`, `continue`, and similar keywords from "less important" keywords such as `let` or `impl`. Thus I added a new syntax type called `rustCtrlFlow` so color scheme authors can specifically target the control flow related keywords.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -24,17 +24,17 @@ syn keyword   rustOperator    as
 
 syn match     rustAssert      "\<assert\(\w\)*!" contained
 syn match     rustPanic       "\<panic\(\w\)*!" contained
-syn keyword   rustKeyword     break
+syn keyword   rustCtrlFlow    break
+syn keyword   rustCtrlFlow    continue
+syn keyword   rustCtrlFlow    return
+syn keyword   rustCtrlFlow    yield
 syn keyword   rustKeyword     box nextgroup=rustBoxPlacement skipwhite skipempty
-syn keyword   rustKeyword     continue
 syn keyword   rustKeyword     crate
 syn keyword   rustKeyword     extern nextgroup=rustExternCrate,rustObsoleteExternMod skipwhite skipempty
 syn keyword   rustKeyword     fn nextgroup=rustFuncName skipwhite skipempty
 syn keyword   rustKeyword     in impl let
 syn keyword   rustKeyword     macro
 syn keyword   rustKeyword     pub nextgroup=rustPubScope skipwhite skipempty
-syn keyword   rustKeyword     return
-syn keyword   rustKeyword     yield
 syn keyword   rustSuper       super
 syn keyword   rustKeyword     where
 syn keyword   rustUnsafeKeyword unsafe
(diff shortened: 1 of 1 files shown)


da-x:
Thanks! It's a good step, and it would be better even to bring us closer to what the docs say in `:help group-name`. We should assign keywords to their respective minor groups. `if` can be under `Conditional`, `loop` can be in the `Repeat` minor group as well, etc.

mandreyel:
@da-x That's an even better idea. I can look into that, this weekend possibly, if that's fine!

da-x:
@mandreyel Thanks!

z2oh:
@mandreyel Are you still planning on looking into this? If not, I'd be willing to take a stab at it.

mandreyel:
I'm so sorry, lots happened recently and I never got around to this! @z2oh I'd still like to finish this, if you don't mind. But I'll let you know if I can't finish it!

mandreyel:
@da-x Alright, I assigned loop keywords to `Repeat`.

I noticed that `rustStructure` and 'rustTypedef` are also assigned to `Keyword` instead of their respective minor groups ('Structure' and 'Typedef', respectively). There are even comments saying that it would be more fitting to do so. Is there a reason for assigning them to `Keyword` instead of their minor groups?

z2oh:
@da-x sorry to bother you... do you have any sort of update on this? I am very much looking forward to improved syntax highlighting 😀

chris-morgan:
Apparently I never ended up actually posting this comment, and so it simply sat in the tab in Firefox for a month or two—

More precise groups exist, but consensus (between myself and whoever wrote the various parts in the first place) was that distinguishing them didn’t look as good; that’s why we left them, so that you *can* `hi link rustStructure Structure`, but that it doesn’t *default* to that.

(However, for reference, I belong to the school of thought that says colour schemes are generally too fancy. A few years ago I threw away my colour schemes and wrote a new one from scratch, which I called “bland”. Keywords are bold, strings are red, comments are italic green, numbers are blue, and that’s about it. I should therefore be considered biased.)

da-x:
Hi, sorry for the late response.

I've done some analysis, and it's hard for me to find reasons to align myself with the previous consensus. 

My rationale is the following: Being that it is up to the color scheme on whether to allow differentiation on keywords types, and then it is up to the user on whether to utilize it in his or her color scheme (as a side note example, my own color scheme does not (yet!) differentiate loop and conditional keywords from other keywords), combined with the fact that many syntax highlight plugins that come built-in with Vim *do* allow for this differentiation, then we should allow for this as well in `rust.vim`.

The main reason is consistency. For example, suppose that I was a using the trio of Lua, Haskell, and Rust in the same project, and have configured differentiation for the `Structure` subgroup in my color scheme, then I'd be surprised to find that Lua and Haskell do mark their 'data structure definition'-related keyword with the Structure keyword, but Rust does not.

```
$ grep -R Structure /usr/share/vim/vim81/syntax  | egrep 'haskell|lua|rust'
./haskell.vim:syn match hsStructure             "\<\(class\|data\|deriving\|instance\|default\|where\)\>"
./haskell.vim:hi def link hsModule                        hsStructure
./haskell.vim:hi def link hsStructure                     Structure
./rust.vim:syn keyword   rustStructure struct enum nextgroup=rustIdentifier skipwhite skipempty
./rust.vim:hi def link rustStructure     Keyword " More precise is Structure
./rust.vim:hi def link rustUnion         rustStructure
./lua.vim:hi def link luaTable          Structure
```

If the original intention for `Structure` was to target C-like structures, it was largely missed by plugin writers. Actually, I like what the other plugin writers did. So it's better to align ourselves to the conventions de-facto and not surprise the users (hence the PR). And on whether it looks good or not, depends largely on the color scheme and the taste of the user. Inconsistency looks worse, IMHO.

So let's continue with the changes?

PR #272, opened 2018-10-16
improve portability of RustTest command

z2oh:
The RustTest command added in #267 is super useful! I updated rust.vim and tried to use it and ran into some issues with my (admittedly unusual) system configuration.

I use `set nomagic` in my `.vimrc`, so the regex to search for `#\[test]` was searching for the literal string `#\[test]`, backslash including. Prefixing the regex with `\m\C` ensures we are using magic and matching case (i.e. ignoring the user's custom `magic` and `ignorecase` settings). This is recommended by [Google's Vimscript style guide](https://google.github.io/styleguide/vimscriptguide.xml?showone=Regular_Expressions#Regular_Expressions).

I also use [fish](https://fishshell.com/) as my shell, which does not support the `&&` operator. I separated the two commands to be on separate lines, which allows fish users to take advantage of this feature. I don't think this should affect users of other shells, though I would appreciate someone verifying this belief.

@rhysd can you check over my changes and make sure I didn't affect any of the functionality? Thank you :)

### DIFF
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -443,12 +443,12 @@ function! s:SearchTestFunctionNameUnderCursor() abort
     let cursor_line = line('.')
 
     " Find #[test] attribute
-    if search('#\[test]', 'bcW') is 0
+    if search('\m\C#\[test\]', 'bcW') is 0
         return ''
     endif
 
     " Move to an opening brace of the test function
-    let test_func_line = search('^\s*fn\s\+\h\w*\s*(.\+{$', 'eW')
+    let test_func_line = search('\m\C^\s*fn\s\+\h\w*\s*(.\+{$', 'eW')
     if test_func_line is 0
         return ''
     endif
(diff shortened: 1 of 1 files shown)


autoload/rust.vim
```diff
@@ -472,7 +472,8 @@ function! rust#Test(all, options) abort
     let pwd = shellescape(pwd)
 
     if a:all
-        execute '!cd ' . pwd . ' && cargo test ' . a:options
+        execute 'silent !cd' pwd
+        execute '!cargo test' a:options
```
rhysd: I think this change does not work since `!cd` just execute `cd` command in sub shell and it does not affect Vim's current directory.

How about this?

```vim
let cd = haslocaldir() ? 'lcd' : 'cd'
let saved = getcwd()
try
    silent execute cd pwd
    execute '!cargo test' a:options
finally
    silent execute cd saved
endtry
```

This changes current working directory temporarily and restores it after.

Please also remove above `let pwd = shellescape(pwd)` line since it is not necessary for `:cd` or `:lcd` command.

autoload/rust.vim
```diff
@@ -485,7 +486,8 @@ function! rust#Test(all, options) abort
             echohl None
             return
         endif
-        execute '!cd ' . pwd . ' && cargo test ' . func_name . ' ' . a:options
+        execute 'silent !cd' pwd
+        execute '!cargo test' func_name a:options
```
rhysd: ditto

rhysd:
@z2oh Thank you for this patch. I did not care about your points. I have reviewed your patch. Could you take a look at my comments above?

z2oh:
@rhysd Thanks for your review! I included your suggested changes in a2e1876.

PR #274, opened 2018-10-24
Use --manifest-path to run tests from outside project directory

rhysd:
Current `:RustTest` implementation works fine. But I noticed that using `--manifest-path` is shorter and safer than changing the working directory temporarily.


da-x:
Thanks!

PR #275, opened 2018-10-27
rustfmt.vim: --write-mode=overwrite is not recognized anymore

SebastianKeller:
rustfmt (stable) does not recognize the option '--write-mode=overwrite'. This change makes :RustFmt work again for stable.

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -56,7 +56,7 @@ function! s:RustfmtWriteMode()
     if g:rustfmt_emit_files
         return "--emit=files"
     else
-        return "--write-mode=overwrite"
+        return ""
     endif
 endfunction
 


da-x:
`rustfmt` version should normally be auto-detected, so it would know whether it is an old version and pass `--emit` instead. What do you get from `:call rustfmt#DetectVersion()` and `:echo g:rustfmt_emit_files`?

SebastianKeller:
`:call rustfmt#DetectVersion()` does not output anything at all and `:echo g:rustfmt_emit_files` prints `1`.

`rustfmt --version => rustfmt 0.99.4-stable (1c408818 2018-08-27)`
`which rustfmt => /Users/SebastianKeller/.cargo/bin/rustfmt`

SebastianKeller:
Ah, nevermind. It seems to work now. I don't know why i had that issue last night then.

PR #282, opened 2018-11-24
Add new keywords

hcpl:
* `async` from https://github.com/rust-lang/rfcs/pull/2394;
* `existential` from https://github.com/rust-lang/rfcs/pull/2071.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -24,10 +24,12 @@ syn keyword   rustOperator    as
 
 syn match     rustAssert      "\<assert\(\w\)*!" contained
 syn match     rustPanic       "\<panic\(\w\)*!" contained
+syn keyword   rustKeyword     async
 syn keyword   rustKeyword     break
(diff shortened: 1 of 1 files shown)


hcpl:
cc @da-x

hcpl:
Hi, @da-x!

> `existential` is a weak keyword, meaning it has special meaning only in specific contexts. In other contexts it can be a valid variable name.

I didn't know this was a thing. I searched for places where "`existential` is a weak keyword" was mentioned in https://github.com/rust-lang/rust and https://github.com/rust-lang/rfcs, but couldn't find any. Could you provide the source?

da-x:
@hcpl: https://github.com/rust-lang/rust/blob/abe19a730576cf7ead1bf7995271b53d551ea37f/src/libsyntax_pos/symbol.rs#L417

da-x:
Thanks!

PR #284, opened 2018-11-29
Add existential as a keyword

dlrobertson:
Highlight existential if followed by a typedef.


dlrobertson:
Just noticed #282... My bad :flushed:

PR #289, opened 2018-12-19
Remove incorrect rustfmt version check

de-vri-es:
There was a version check in `rusfmt.vim` to determine if `--unstable-features` should be passed to `rustfmt` or not. This PR removes that check. The check seems to assume that with 1.0 the `--file-lines` feature is stabilized, but it is not:

```
$ rustfmt -V
rustfmt 1.0.1-nightly (be13559 2018-12-10)
$ rustfmt --file-lines ''
Unstable option (`--file-lines`) used without `--unstable-features`
```

Additionally, this PR replaces the rather odd `1 - (foo !~# bar)` pattern with simply `foo =~# bar`. I asked around at #vim on freenode, nobody could give me a valid reason for the former.

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -22,7 +22,7 @@ endif
 function! rustfmt#DetectVersion()
     " Save rustfmt '--help' for feature inspection
     silent let s:rustfmt_help = system(g:rustfmt_command . " --help")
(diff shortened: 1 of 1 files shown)


da-x:
Thanks!

de-vri-es:
Thanks for the quick review! :)

PR #290, opened 2018-12-20
Improve 'async' contextual keyword highlighting

rhysd:
`async` is a contextual keyword. For example, when it is used in context of scope resolution (`::`), `async` can be used as module name. In the case, the `async` name should be highlighted as normal identifier, not a keyword.

This patch restricts highlighting `async` as keyword only when space or newline is following. So now `async` in module resolution such as `foo::async::X` is highlighted as normal module name.

Here is a screenshot:

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -26,7 +26,7 @@ syn match rustExistentialContextual /\<existential\_s\+type/ transparent contain
 
 syn match     rustAssert      "\<assert\(\w\)*!" contained
 syn match     rustPanic       "\<panic\(\w\)*!" contained
-syn keyword   rustKeyword     async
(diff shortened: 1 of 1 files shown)


da-x:
[According to the docs](https://doc.rust-lang.org/reference/keywords.html) `async` is supposed to be a strict keyword, but only since edition 2018. We have the following issue for the renaming of `async` module in request: https://github.com/seanmonstar/reqwest/issues/357 to be compatible with edition 2018.

Until edition 2018 is default in Rust and Cargo, we should default to highlighting like it's edition 2015, and in the future, auto-detect the currently configured edition in Cargo.toml and perform syntax highlighting accordingly. For now, a bridging solution that handles the contextual use of `async` suffices.

Thanks!

PR #291, opened 2018-12-22
Use :terminal for running tests on :RustTest

rhysd:
Currently `:RustTest` runs tests with `:!`. However, it uses entire screen. We cannot check the results and Vim buffer at the same time.

Neovim and Vim 8+ now has a terminal and `:terminal` can start a command in the terminal. I prefer it to `:!`.

This patch uses `:terminal` for running `cargo test` if possible.

### DIFF
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -468,10 +468,20 @@ function! rust#Test(all, options) abort
     if manifest ==# ''
         return rust#Run(1, '--test ' . a:options)
     endif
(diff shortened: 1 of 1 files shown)


z2oh:
This is very cool, though I am having trouble finding documentation on the `:terminal` command. My colors in this window are incorrect; I know this is not a forum for vim support but would you happen to know why this is happening @rhysd?

Screenshot (notice the background color for text in the top pane):
screen shot 2019-01-10 at 15 03 02

PR #292, opened 2018-12-22
RustTest: specify module name to run exact one test case

rhysd:
I found one issue for `:RustTest`.

- `some_test.rs`

```
#[test]
fn test_foo() {}
```
- `another_test.rs`

```
#[test]
fn test_foo() {}
```

Running `:RustTest` in `test_foo()` function in `some_test.rs` runs both test cases `some_test::test_foo` and `another_test::test_foo`.

To solve this issue, this patch extracts module name from its file name and specify `mod_name::test_func` instead of `test_func`.


da-x:
I wish there was something more precise for this. Opened: https://github.com/rust-lang/cargo/issues/6478

rhysd:
Looks good. I agree that it's definitely better that `cargo` supports running a test specified by position in file.
Until the feature is supported, I want to add this patch as the best effort.

PR #294, opened 2018-12-27
Set `include` and improve `includeexpr`

dimbleby:
Improved cross-file search:

- recognise `use crate::foo::bar::baz` as being an 'include' statement
- enhance `includeexpr` to handle 2018 edition, among other things

it's a bit unfortunate that `isfname` is a global variable in vim; I've followed the lead of the perl plugin, [here](https://github.com/vim/vim/blob/6436cd83f90a0efc326798792e49e8ff96a43dce/runtime/ftplugin/perl.vim#L38-L43) in setting it anyway.

(I suppose most people use RLS for this sort of thing, and that is mostly superior.  But maybe that's not always available for one reason or another; and anyway, I see no reason not to let Vim do its best).

### DIFF
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -3,23 +3,59 @@
 " Last Modified: May 27, 2014
 " For bugs, patches and license go to https://github.com/rust-lang/rust.vim
 
-" Jump {{{1
-
-function! rust#Load() 
+function! rust#Load()
     " Utility call to get this script loaded, for debugging
 endfunction
 
 function! rust#GetConfigVar(name, default)
     " Local buffer variable with same name takes predeence over global
-    if has_key(b:, a:name) 
+    if has_key(b:, a:name)
         return get(b:, a:name)
     endif
-    if has_key(g:, a:name) 
+    if has_key(g:, a:name)
         return get(g:, a:name)
     endif
     return a:default
 endfunction
 
+" Include expression {{{1
+
+function! rust#IncludeExpr(fname) abort
+    " Remove leading 'crate::' to deal with 2018 edition style 'use'
+    " statements
+    let l:fname = substitute(a:fname, '^crate::', '', '')
+
+    " Remove trailing colons arising from lines like
+    "
+    "     use foo::{Bar, Baz};
+    let l:fname = substitute(l:fname, ':\+$', '', '')
+
+    " Replace '::' with '/'
+    let l:fname = substitute(l:fname, '::', '/', 'g')
+
+    " When we have
+    "
+    "    use foo::bar::baz;
+    "
+    " we can't tell whether baz is a module or a function; and we can't tell
+    " which modules correspond to files.
+    "
+    " So we work our way up, trying
+    "
+    "     foo/bar/baz.rs
+    "     foo/bar.rs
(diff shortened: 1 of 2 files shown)


da-x:
Hi @dimbleby, thanks for your contribution. Sorry for the delayed response.

I don't feel quite comfortable with the change to the global `isfname` option.
Instead of overriding `isfname`, let's try to follow from the suggestion given in the help for `include`:

	Normally the 'isfname' option is used to recognize the file name that
	comes after the matched pattern.  But if "\zs" appears in the pattern
	then the text matched from "\zs" to the end, or until "\ze" if it
	appears, is used as the file name.  Use this to include characters
	that are not in 'isfname', such as a space.  You can then use
	'includeexpr' to process the matched text.
	See |option-backslash| about including spaces and backslashes.

dimbleby:
Ah, that is better, thanks.  More elaborate regex in the most recent commit...

dimbleby:
Although setting `include` this way allows `[i` and suchlike to work without having to touch `isfname`, it doesn't do anything for `gf`.  

That is, if you're sitting on a `use crate::foo::bar::baz` line now and `gf`, you'll end up looking for `crate`, or `foo`, or `bar`, or `baz`, depending on where the cursor is - whereas if we added `:` to `isfname` then we would look for `foo/bar/baz.rs` and so on per the `includeexpr`.

I think that the version as currently submitted (don't touch `isfname`, give up that usage of `gf`) is better than having filetype-specific code touching global options, but I wanted to spell out the trade that is being made.

da-x:
If `gf` is an the main issue, how about we have an Rust-dedicated implementation of it?

Like here:

https://github.com/amix/vimrc/blob/993ed5530408f0d737ee37b152c0e6c0cd1211ef/sources_non_forked/open_file_under_cursor.vim/plugin/open_file_under_cursor.vim

(But with a change where `set isfname+=:` is done locally inside the function and immediately reverted if it performed any change. Also, Vim lets you do `setlocal isfname`, to the effect that it is a global set now, to be future-compatible to some degree should they decide to make this setting buffer-local.)

I think it is called for, because these days you can have even more complex `use` statements, e.g. with nesting:

```rust
use crate::x {
    y::b {
    },
    z
}
```

Though it requires more complex parsing, but we can leave it as a future To-Do, for a future PR. Also as a future To-Do, replace `findfile` with exact filename lookup relative to the nearest `Cargo.toml` so workspaces are handled properly and `target/` is not accidentally searched. And finally as a To-Do, extend the logic to handle built-in modules `mod name {}` where the looked upon files don't exist. But more hopefully, in the future, RLS should allow us to "jump-to-definition" when hovering over 'use` declrs and not re-implement all this logic ourselves.

dimbleby:
I think I'm content to leave this pull request as it stands - I see it as a clear and significant improvement on what we have today, and I'm not personally all that bothered about `gf`. 

(If you or some future A. N. Other wants to make `gf` work then sure, the approach you propose sounds plausible.  I also agree that it's undesirable to get too elaborate: RLS should be what gives us really accurate goto-definition etc, and I'm fine with the Vim plugin just including some sensible heuristics that work out pretty often).

PR #296, opened 2018-12-28
better path

dimbleby:
following #294
- that one has us recognise and build filenames more accurately
- this one has us look for those files in a sensible place, based on finding `Cargo.toml`


dimbleby:
(rebased to fix merge conflict)

PR #300, opened 2019-01-09
Fix auto format on save don't work when switch buffer

dbger:
Produce this issue:

```
1) open main.rs
2) :e new.rs 
3) :bp switch back main.rs
4) edit and :w, auto format don't work
```


da-x:
Performed an equivalent fix in master. Thanks

PR #301, opened 2019-01-10
Use 'setfiletype' to avoid load 'ftplugin/rust.vim' twice 

dbger:
`au BufNewFile,BufRead *.rs setf rust` in $VIMRUNTIME/filetype.vim had set the rust filetye,  `autocmd BufRead,BufNewFile *.rs set filetype=rust` will overrule it. 
This cause ftplugin/rust.vim will be loaded twice when the first is not loaded yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,11 @@ If you set g:rust_clip_command RustPlay will copy the url to the clipboard.
 
         let g:rust_clip_command = 'xclip -selection clipboard'
 
+### Running a test under cursor
+
+In cargo project, the `:RustTest` command will run a test under the cursor.
+This is useful when your project is bigger and running all tests take longer time.
+
 ## Help
 
 Further help can be found in the documentation with `:Helptags` then `:help rust`.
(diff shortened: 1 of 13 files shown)


chris-morgan:
I am startled to see `set filetype=` rather than `setf `, because the latter is definitely the right way to do it; looks like it was added the wrong way right back at the start, seven years ago: 7d6eef95d0d0c3c84c838945283d13bf976ed907.

I don’t think I’ve ever seen `setfiletype` spelled in full before. `setf` is definitely the idiomatic spelling, change it to that then I’ll give r+.

dbger:
@chris-morgan It's done.

dbger:
> 
> 
> I am startled to see `set filetype=` rather than `setf `, because the latter is definitely the right way to do it; looks like it was added the wrong way right back at the start, seven years ago: [7d6eef9](https://github.com/rust-lang/rust.vim/commit/7d6eef95d0d0c3c84c838945283d13bf976ed907).
> 
> I don’t think I’ve ever seen `setfiletype` spelled in full before. `setf` is definitely the idiomatic spelling, change it to that then I’ll give r+.

It's done.

PR #302, opened 2019-01-14
Update `ctags` regexes to include `pub` modifiers

ErichDonGubler:


chris-morgan:
Good catch, crate visibility does need to be added. The regular expression chosen needs work, however:

- No space is required after the `)`, so `pub(crate)fn foo(){}` is legal.
- Spaces inside the parens are fine too.
- You can have a visibility spec `in ‹path›`, so adding whitespace in there and `:` for the path is necessary.

All up, I’m thinking `(pub([ \t]*\([a-zA-Z0-9_ \t:]+\)[ \t]*|[ \t]+))`.

ErichDonGubler:
Okay. I'll be more than happy to take a look and make fixes -- I'll probably get to this sometime next week, though.

PR #308, opened 2019-02-16
Update the documentation to clarify Vim 8 requirement

zdb999:
Clarifies that Vim 8 is necessary for the plugin to work correctly. When installing, I spent several hours trying figure out why certain commands weren't working. It turns out the problem was that I was using Vim 7.1, but I couldn't find the version 8 requirement anywhere in the documentation. These small changes clarify the issue so others don't have this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ## Description
 
 This is a Vim plugin that provides [Rust][r] file detection, syntax highlighting, formatting,
-[Syntastic][syn] integration, and more.
+[Syntastic][syn] integration, and more. It requires Vim 8 or higher.
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 This is a Vim plugin that provides [Rust][r] file detection, syntax highlighting, formatting,
 [Syntastic][syn] integration, and more. It requires Vim 8 or higher for full functionality.
-Some things man not work on earlier versions. 
+Some things may not work on earlier versions. 
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -3,7 +3,8 @@
 ## Description
 
 This is a Vim plugin that provides [Rust][r] file detection, syntax highlighting, formatting,
-[Syntastic][syn] integration, and more. It requires Vim 8 or higher.
+[Syntastic][syn] integration, and more. It requires Vim 8 or higher for full functionality.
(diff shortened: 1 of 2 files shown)


chris-morgan:
This crate is expected to work back to at least Vim 7.3—there are particular adjustments in behaviour for 7.3.541 and 7.4.355, for example. There will be minor degradation in some areas such as indentation. I don’t actually know how the Rust* commands work and what their compatibility may be, I haven’t paid much attention to those.

7.1 is positively ancient. 7.2 came out 10½ years ago! I have no idea what could be broken in it, and definitely have no interest in fixing it. Vim 8 has good stuff, and I wouldn’t actually mind drawing a line in the sand and saying “if you use less than Vim 8.0.0, use such-and-such a hash of this repository and you won’t be getting any more updates”.

da-x:
It's likely that we may still need to support some versions of Vim 7.x. Systems like CentOS 7.x, which are still in active maintenance, are going to be supported years ahead, and these still come with Vim 7.4 pre-installed. People are using these systems as development environments. I think that there are some small fixes that can be made for 7.4 - I am not sure about 7.1 though.

I agree that we can have features disabled based on detection of an older version. I suggest the following - perhaps change the wording in this PR so that "for full functionality Vim 8 is required. For earlier versions some functionality will not work".

Also, it's understandable that it should not break with unpleasant errors, and that it should convey about the disabled features in a useful manner. So, maybe as an extension of this PR, we should identify the commands for which we are not planning to add support for older versions of Vim, and emit a nice error message upon their invocation with those versions?

zdb999:
I'm using Ubuntu 14.04, and Vim 7.1 was the built in version in the official apt-get repository. Ubuntu packages are often extremely out of date. The problem is even worse on older LTS editions, but the stability is great and I haven't bothered to upgrade.

Prior to updating, the only commands that worked were the Cargo shortcuts. RustFmt wouldn't work, and feature highlighting and tagbar integration both failed. Additionally, the RustInfo command was not recognized, which was particularly frustrating as it led me to believe that the plugin hadn't downloaded to the right place. (I was using the Plug package manager.) Providing some useful information on older versions with that command would be helpful; a simple message to upgrade to a newer Vim would have saved me a lot of time.

zdb999:
Also, I changed the wording to align with @da-x's suggestion.

da-x:
Thanks!

PR #309, opened 2019-02-24
Include entries starting with ":::" in quickfix window.

jmcomets:
Currently, the `errorformat` does not match additional files mentioned in a
compilation error.

In the following example, only the first file mentioned is matched (`src/node4.rs`, line 118, column 1):

```
error[E0046]: not all trait items implemented, missing: `children`
   --> src/node4.rs:118:1
    |
118 | impl<'a, T> NodeImpl<'a, T> for Node4<'a, T> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `children` in implementation
    |
   ::: src/lib.rs:134:5
    |
134 |     fn children(&self) -> Box<dyn Iterator<Item=(u8, &'a Child<'a, T>)> + 'a>;
    |     -------------------------------------------------------------------------- `children` from trait
```

This pull-request will include the second file mentioned (`src/lib.rs`, line 134, column 5).

### DIFF
--- a/compiler/rustc.vim
+++ b/compiler/rustc.vim
@@ -34,6 +34,7 @@ CompilerSet errorformat=
             \%Wwarning:\ %m,
             \%Inote:\ %m,
             \%C\ %#-->\ %f:%l:%c,
+            \%C\ %#:::\ %f:%l:%c,
             \%E\ \ left:%m,%C\ right:%m\ %f:%l:%c,%Z
 
 " Old errorformat (before nightly 2016/08/10)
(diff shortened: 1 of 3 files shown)


jmcomets:
FYI I'm still working on things, I'll update this PR when things are ready.

jmcomets:
After extensive testing of Vim's `errorformat` help, I've decided to give up. Continuation patterns can only specify the file via `%f` ONCE per error. Which is the underlying problem. The other solution would be to match each file in a compilation stack as its own error/warning, which doesn't make sense...

PR #310, opened 2019-02-27
Fix syntax for `as` keyword

cloudhead:
`as` is a keyword, not an operator.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -20,7 +20,6 @@ syn keyword   rustTypedef type nextgroup=rustIdentifier skipwhite skipempty
 syn keyword   rustStructure struct enum nextgroup=rustIdentifier skipwhite skipempty
 syn keyword   rustUnion union nextgroup=rustIdentifier skipwhite skipempty contained
 syn match rustUnionContextual /\<union\_s\+\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*/ transparent contains=rustUnion
-syn keyword   rustOperator    as
 syn keyword   rustExistential existential nextgroup=rustTypedef skipwhite skipempty contained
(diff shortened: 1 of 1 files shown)


chris-morgan:
Sure, `as` is a keyword inasmuch as it’s made of letters rather than symbols, but it’s a binary operator, with an expression as its first operand and a type as its second.

I won’t outright close this, but I’m very strongly disinclined to accept the change, and good arguments will be required to sway me.

cloudhead:
I'm not making this up - it's what the various references say: https://doc.rust-lang.org/reference/keywords.html

chris-morgan:
I repeat the first paragraph of my last comment.

cloudhead:
The rust specification considers it a keyword, regardless of whether it's made of letters or symbols, or acts as an operator. Semantically it may work like an operator, but syntactically it's a keyword.

chris-morgan:
The two are not mutually exclusive. Also from the Rust Reference, [`as` is classed an operator](https://doc.rust-lang.org/reference/expressions/operator-expr.html#type-cast-expressions) (and this bit has been in there from the start):

> A type cast expression is denoted with the binary operator `as`.

With this concrete source, I am now strongly inclined to close this, but I’ll leave it open a bit longer in case anyone else wishes to weigh in, or in case you want to replace it with a different commit in the following way (or create a new PR if that’s easier for you):

If you would strongly prefer keyword highlighting for it, I’d accept a new syntax group `rustAs` which defaulted to `rustOperator`, which would allow you to readily override it for yourself.

cloudhead:
> The two are not mutually exclusive.

Indeed - it depends how you want to treat it in the syntax file. If you look at the example you linked, `as` is highlighted the same as keywords (`let`, `fn`), not operators (which don't get highlighted) - so whatever highlighter `doc.rust-lang.org` uses, treats it as a keyword too.

Feel free to close if you think this isn't a good enough reason - I don't have anything else to add - I don't think adding a `rustAs` makes that much sense either (seems ugly) - I'll just add an override locally in `after/syntax/rust.vim` :+1:

PR #311, opened 2019-03-05
ftdetect: Use `setf FALLBACK` to set Cargo.toml filetype

heftig:
[vim-toml recently started using `setf`.][1] As a result, the filetype set by rust.vim is no longer overridden. 
Use `setf FALLBACK` as this is the proper mechanism for uncertain filetypes.

[1]: https://github.com/cespare/vim-toml/commit/d257faaa32ffee1794038f2342ba315760c33fc2

### DIFF
--- a/ftdetect/rust.vim
+++ b/ftdetect/rust.vim
@@ -1,6 +1,6 @@
 " vint: -ProhibitAutocmdWithNoGroup
 
 autocmd BufRead,BufNewFile *.rs setf rust
-autocmd BufRead,BufNewFile Cargo.toml if &filetype == "" | set filetype=cfg | endif
+autocmd BufRead,BufNewFile Cargo.toml setf FALLBACK cfg
 
 " vim: set et sw=4 sts=4 ts=8:


chris-morgan:
By the looks of it, FALLBACK only landed in [Vim 8.0.0613](https://github.com/vim/vim/commit/3e54569b17683318e0cb6693ab0024c2ad1e3e8f), so we definitely can’t depend on its presence. It makes sense to use it, but you’ll need a fallback for it where it’s not supported.

heftig:
I'm not sure there's a viable fallback for old Vims, then. 

I suppose I could condition it on the Vim version, but honestly I would rather eliminate the line completely. The `cfg` syntax is not only an incomplete highlight for TOML but also a misleading one. It gets confused when inline tables or arrays are involved and is only acceptable for trivial files.

What would you accept? Right now it's broken and doesn't let vim-toml apply it's syntax.

chris-morgan:
I think breaking Cargo.toml highlighting for people that update rust.vim, don’t have vim-toml and have not-the-latest Vim is probably acceptable. That’d suggest `silent! setf FALLBACK cfg`.

… but hmm, on current Vim `setf FOO cfg` already silently does nothing; and I don’t have an old Vim handy any more to readily check if `setf FALLBACK cfg` already safely does nothing without needing `silent!`.

heftig:
Seems that Vim rejects filetypes with spaces in them since v8.0.0056.

heftig:
`setf FALLBACK cfg` fails silently as far back as Ubuntu 12.04's `2:7.3.429-2ubuntu2.2`. I've tried to run older Docker containers like `ubuntu:10.04` and `debian:7`, but they're all defunct as apt's http backend segfaults.

heftig:
Can this be merged? As I explained above, in all versions I tested, if `FALLBACK` is not supported, the command will silently do nothing, so it's safe.

da-x:
Confirmed with 7.4.1099.

Thanks!

PR #316, opened 2019-04-09
Restore cwd on rustfmt failure before opening lwindow

derwiath:
If rustfmt fails RunRustFmt helpfully opens the lwindow to show
the errors rustfmt encountered.
The 'lwindow' command opens the location list window, and makes that
the active vim window.

Before rustfmt is executed RunRustFmt change CWD to the directory
where the file to be formated resides. CWD is restored before exiting
RunRustFmt.

The problem was that location list window was opened before restoring
CWD, which meant that we only restored the directory of the lwindow.
The window holding the formatted rust file is left with CWD of that
file.

The obvious solution is to restore the CWD *before* opening the
location list window.

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -145,6 +145,7 @@ function! s:RunRustfmt(command, tmpname, fail_silently)
 
     call delete(l:stderr_tmpname)
 
+    let l:open_lwindow = 0
     if v:shell_error == 0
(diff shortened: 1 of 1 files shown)


da-x:
Nice catch! I confirmed the fix with a plain configuration. 

The original issue is invisible for users that have scripts that automatically set buffer current directory to be project-relative.

Issue #318, opened 2019-04-28
Plugin appears not to get loaded by neovim


dasteihn:
rust.vim version: 

The issue appears to be very similar to issue 288, though I don't use the vim-css Plugin like the other guy did.

I use vimplug and rust.vim commit 9984b97
Editor is Neovim, version 0.1.7

Steps to reproduce:
Install the rust.vim and open a Rust-file

Expected vs. actual behavior:
It's expected to have working syntax highlighting. Instead, all that is highlighted for me are the brackets, everything else is treated as normal text. 

Paste debugging info from the Rust Vim plugin via _one_ of the following
commands: `:RustInfo`,  `:RustInfoToClipboard`, or `:RustInfoToFile <filename>`.

The `RustInfo`(nor similar) command appears not to be found by nvim. Yet the PlugStatus indicates that the plugin is ready to use
```
Finished. 0 error(s).
[===]

- nerdcommenter: OK
- nerdtree: OK (not loaded)
- rust.vim: OK
```

Vimrc:
```
:set number
:colorscheme torte
:set colorcolumn=81,91
:set tabstop=4
:set smarttab
:set autoindent
:set shiftwidth=4
let g:airline_powerline_fonts = 0

let mapleader = ","

call plug#begin('~/.nvim/plugged')
Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }
Plug 'scrooloose/nerdcommenter'
Plug 'rust-lang/rust.vim'
call plug#end()
```

mhinz:
1. Nvim v0.1.7 is pretty old. There were more than 5000 patches since then.
1. Is `~/.nvim` a symlink? Because it's not a directory that Nvim checks by default. What `~/.vim` is to Vim, is `~/.config/nvim` to Nvim.
1. Use `:scriptnames` to see exactly which files got sourced.

dasteihn:
1. That's the thing  with Debian, you have age-old stuff. The awkward thing is that the plugin stopped working from one week to another.
2. No, that's a directory. But as most installation guides portray it it's the script file destination place for plugins
3. rust.vim gets loaded

It's notable that the other plugins work, so I'm not sure if there is a problem with the plugin folders

mhinz:
`:set filetype syntax` shows `rust` in both cases?

dasteihn:
No, filetype appears to be something called "hercules" for `.rs` files

[...]

Issue #306, opened 2019-02-08
Automatic syntax highlighting with .rs files does not work


alkorang:
**rust.vim version:**
commit 12f549f9e4939bca53c8a87e6921a36fb143af9a

**Steps to reproduce:**

On CentOS 7, install vim 7.4
```
# yum install vim-enhance vim-common
```

Then, install rust.vim using vim-plug
```
$ curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
    https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
```
`~/.vimrc` file:
```
call plug#begin('~/.vim/plugged')
Plug 'rust-lang/rust.vim', { 'tag' : '12f549f9e4939bca53c8a87e6921a36fb143af9a' }
call plug#end()
```

After saved `~/.vimrc`, run `:PlugInstall` in vim or `$ vim +PlugInstall +qall` on terminal.

**Expected vs. actual behavior:**

* Expected:
When I opened rust source file with .rs extension, for example `main.rs`, then vim automatically highlights Rust syntax.
(And this is the actual behavior until commit 329fa788f56977d73ff958388b8e08ea7bb885f3.)
![image](https://user-images.githubusercontent.com/11420879/52477963-1ddd6880-2be7-11e9-9ec6-0ac795cfe5e6.png)

* Actual:
With commit 12f549f9e4939bca53c8a87e6921a36fb143af9a, vim automatically highlights syntax but it is not for Rust.
![image](https://user-images.githubusercontent.com/11420879/52477998-41081800-2be7-11e9-8cd9-287c5a68ca83.png)
When I run `:set syntax=rust` manually, then Rust syntax highlight works.

I cannot get additional debug information because `:RustInfo` failed with E492 error.
![image](https://user-images.githubusercontent.com/11420879/52478514-ecfe3300-2be8-11e9-81c8-4acc593e9345.png)

chris-morgan:
I can’t reproduce this on Vim 8.1.1 on Windows, probably for the reasons mentioned below.

This smells like another ftdetect rule butting in first, thereby triggering exactly the deliberate and intended difference between `setfiletype rust` and `set filetype=rust`.

Please run this to determine what the filetype actually *is* on a .rs file, and where it was last set:

```
:verbose set filetype?
```

[...]

PR #319, opened 2019-04-28
Improve filetype setting

mhinz:
- `:set filetype` forces a filetype.
- `:setfiletype` only sets a filetype if it didn't already happen earlier due to
the same event.

Current versions of Vim and Neovim have a `$VIMRUNTIME/filetype.vim` that sets
the `rust` filetype for all files with an `.rs` extension.

This plugin did that with `set filetype=rust` as well. But now, with the
filetype getting set twice, `ftplugin/rust/*` was sourced twice as well.

This got fixed in https://github.com/rust-lang/rust.vim/pull/301 by changing
`set filetype=rust` to `setfiletype rust`.

But that fix introduced a regression for all older Vim versions. There is still
a lot of Vim 7.4 around and Debian stable still provides Nvim 0.1.7, both being
very old versions.

The `$VIMRUNTIME/filetype.vim` of these old versions set the filetype to
`hercules` for all `.rs` files via `setfiletype hercules`.

Now, usually (it also depends on how plugins gets sourced) `filetype.vim` gets
sourced _before_ any `ftdetect/*` files from plugins. And since autocmds are
processed in order, Vim would first do `setfiletype hercules` and so any
subsequent `setfiletype rust` fails.

The obvious solution: Force the `rust` filetype for all `.rs` extensions unless
the filetype is already set to `rust`.

So, we have the filetype setting working again for older versions and also avoid
setting the filetype twice.

Fixes https://github.com/rust-lang/rust.vim/issues/306
Fixes https://github.com/rust-lang/rust.vim/issues/318

### DIFF
--- a/ftdetect/rust.vim
+++ b/ftdetect/rust.vim
@@ -1,6 +1,12 @@
 " vint: -ProhibitAutocmdWithNoGroup
 
-autocmd BufRead,BufNewFile *.rs setf rust
+autocmd BufRead,BufNewFile *.rs call s:set_rust_filetype()
 autocmd BufRead,BufNewFile Cargo.toml setf FALLBACK cfg
+
+function! s:set_rust_filetype() abort
+    if &filetype !=# 'rust'
+        set filetype=rust
+    endif
+endfunction
 
 " vim: set et sw=4 sts=4 ts=8:


da-x:
Thanks! It reminds me we should have some automated regression testing. At least the basic plugin functionality should work.

Issue #320, opened 2019-04-29
Syntax highlighting for attributes ends early.


flowbish:
As of Rust 1.34.0, arbitrary syntax is allowed inside of attributes. The syntax highlighting for attributes ends early if square brackets are contained within an attribute.

* rust.vim version: checkout at commit 9984b97

Steps to reproduce:

```Rust
#[some_attribute(
    a = [1, 2],
    b = [3, 4],
)]
...
```

Expected vs. actual behavior:

I expect the entire attribute to be highlighted, but the highlighting ends at the first closing brace (right after `a = [1, 2]`), leaving the rest of the attribute unhighlighted.

PR #321, opened 2019-04-30
Syntax: adjust attribute highlight for Rust 1.34

mhinz:
This is now allowed as well:

    #[some_attribute(
        a = [1, 2],
        b = [3, 4],
    )]

Assumption: The closing ] of an attribute always ends the current line.

Fixes https://github.com/rust-lang/rust.vim/issues/320


chris-morgan:
r-, there’s no reason why `]` must be at the end of the line. Writing attributes on the same line as the item is not uncommon, e.g. `#[inline] fn foo() {}`.

The proper fix here is to make a region for `[` to `]` to ensure balance, as was done with rustBoxPlacementBalance (which looks to be a tad broken, incidentally).

mhinz:
> Writing attributes on the same line as the item is not uncommon, e.g. #[inline] fn foo() {}.

Oh, okay. That makes it a tad more complicated.

When I find a good fix, I'll open another PR.

PR #322, opened 2019-05-05
Allow attributes to contain nested brackets.

### LINKED ISSUE #320: Syntax highlighting for attributes ends early.

flowbish:
With Rust 1.34, proc-macro attributes now accept arbitrary tokens as arguments, allowing a user to write an attribute e.g. `#[my_attribute(a = [1, 2, 3], b = ("x", "y", "z")]`
The previous attribute highlighter would stop at the first `]`, leaving a lot of the attribute unhighlighted. This patch changes the highlighter to be more robust to nested brackets and parentheses.

As before, only derives, comments, and strings get special formatting inside of an attribute. It may make more sense to allow arbitrary syntax highlighting inside of the attribute arguments, but that's outside the scope of this patch.

Fixes #320

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -149,7 +149,12 @@ syn region    rustString      start=+b"+ skip=+\\\\\|\\"+ end=+"+ contains=rustE
 syn region    rustString      start=+"+ skip=+\\\\\|\\"+ end=+"+ contains=rustEscape,rustEscapeUnicode,rustEscapeError,rustStringContinuation,@Spell
 syn region    rustString      start='b\?r\z(#*\)"' end='"\z1' contains=@Spell
 
-syn region    rustAttribute   start="#!\?\[" end="\]" contains=rustString,rustDerive,rustCommentLine,rustCommentBlock,rustCommentLineDocError,rustCommentBlockDocError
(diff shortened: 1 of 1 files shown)


flowbish:
It's been about a month, what can I do to push this PR forward?

da-x:
Looks good to me. If needs refinements, we can do them later.

chris-morgan:
I’m confused: I’m *sure* I wrote a comment about this pointing out the problem, but none is apparent here, so—

This is insufficient because it only handles `(`, and not `[` or `{`. Remember that attributes handle arbitrary token trees.

da-x:
@chris-morgan Thanks - added de998a2014c997d6f8f749a9be195f1cd51aa9c0 in master to address this.

Issue #287, opened 2018-12-10
Messes up '.


krzkrzkrz:
Every time a save is done. It messages up `'.`

As in. In Vim. When I do `'.`, Vim always takes the cursor to the end of the file, instead of the last changed mark

I think it has something to do with `rustfmt`

Is this the expected behaviour or is there an option I need to set in `.vimrc`?

da-x:
Preserving marks is indeed an issue when reformatting replaces the entire buffer content. Perhaps, if not much as changed during the reformatting, we can diff the before-and-after versions and adjust the mark accordingly. I don't see handling for this in other formatters like [neoformat](https://github.com/sbdchd/neoformat), though.

krzkrzkrz:
Remembering/retaining marks would be a good addition

da-x:
I agree, though adjusting the marks on the correct line post-formatting would not be exact without knowing precisely how tokens have shifted during the reformatting. Only rustfmt retains that exact information.

Checking with @nrc - perhaps rustfmt can be given an optional list of `(line, column)` positions, and then it would emit the new positions after formatting? Something like, `--marks-in [filename]`, `--marks-out [filename]`, where file format would be a YAML serialization of `[(String, u64, u64)]`. But perhaps we can come up with a better protocol.

In the future, this will also be helpful in adjusting not only marks, but buffer highlights too.

da-x:
Move to end of file is fixed now by #419 perhaps, want to check?

benknoble:
Could we use `:keepmarks` on the vim sides? It currently only has effect for `:[range]!`, according to the docs.

PR #323, opened 2019-05-19
Try moving the cursor back after running rustfmt.

yangby-cryptape:
I set `let g:rustfmt_autosave = 1`, so the cursor moves to the end of file after I saved it every time.

I'm tired to do move the cursor back manually.
Temporarily, I add this trigger to `autocmd BufWritePre` and `autocmd BufWritePost`.

But, I think add this trigger to the function which moves the the cursor is a better choice.

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -111,6 +111,7 @@ function! s:DeleteLines(start, end) abort
 endfunction
 
 function! s:RunRustfmt(command, tmpname, fail_silently)
+    let l:current_position = getpos('.')
     mkview!
 
     let l:stderr_tmpname = tempname()
@@ -216,6 +217,7 @@ function! s:RunRustfmt(command, tmpname, fail_silently)
     endif
 
     silent! loadview
+    call setpos('.', l:current_position)
 endfunction
 
 function! rustfmt#FormatRange(line1, line2)


da-x:
Related to https://github.com/rust-lang/rust.vim/issues/287

jyc:
This would be useful for me. Right now I can't use RustFmt because it resets my scroll position every time it runs.

da-x:
Hi!

The `mkview` / `loadview` combo should already restore cursor position, and it does so without these changes, at least for me. So we are probably missing something. What do you currently have in `autocmd BufWritePre` and `autocmd BufWritePost`?

JeanMertz:
I could be mistaken, but I think there's still some odd behaviour with this plugin's implementation of `RustFmt`, compared to, for example, Ale.

I used to use Ale before, and it never messed with my undo history, but I recently started using `RustFmt` from this plugin, and I can see some odd behaviour with undo history being gone after a format.

Unfortunately, I don't have a minimal reproducible example right now, but I wanted to point this out nonetheless. I'll be switching back to Ale sometime this week, and hopefully I can get something reproducible to show the difference.

yangby-cryptape:
I almost forgot this PR.

Since I added a trigger in `autocmd BufWritePre` and `autocmd BufWritePost` to reset the cursor, I have never seen this again.
So I don't have a minimal reproducible example now.

PR #327, opened 2019-05-30
Add .await syntax highlighting

seanmonstar:
Also puts `async` into its own group in case one wants to highlight it differently.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -26,7 +26,7 @@ syn match rustExistentialContextual /\<existential\_s\+type/ transparent contain
 
 syn match     rustAssert      "\<assert\(\w\)*!" contained
 syn match     rustPanic       "\<panic\(\w\)*!" contained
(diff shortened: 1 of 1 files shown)


seanmonstar:
@chris-morgan ping :)

chris-morgan:
I’m not convinced about requiring (tho not highlighting) the dot, rather than just highlighting `await` as a keyword. Convoluted code written by madmen that like breaking syntax highlighters:

```rust
a .
await
./**/await
```

Do you have a definite opinion on the presence of the dot in the highlighting, or did that just seem like the thing to do?

seanmonstar:
I had a small test file when I was writing this, and now I cannot remember why I thought it should start with a `.`. I'll remove that, for the sake of the madmen.

seanmonstar:
All set :D

seanmonstar:
@chris-morgan any other thoughts?

chris-morgan:
Thanks for the reminder!

PR #328, opened 2019-06-03
Don't change user's formatoptions

Thomasdezeeuw:
I've hit a bug in which hitting enter it will insert a comment start, any word that overflowed the previous line and set the cursor to the start of the line (before the comment start).

Which result in some amazing stuff like:

```rust 
// Some awesome long
ent. // comm
```

(It should say "Some awesome long comment.")

This change remove setting the formatoptions for the user, if you use Vim you can decide this for yourself.

### DIFF
--- a/ftplugin/rust.vim
+++ b/ftplugin/rust.vim
@@ -38,9 +38,6 @@ else
     setlocal comments=s0:/*!,m:\ ,ex:*/,s1:/*,mb:*,ex:*/,:///,://!,://
 endif
 setlocal commentstring=//%s
-setlocal formatoptions-=t formatoptions+=croqnl
-" j was only added in 7.3.541, so stop complaints about its nonexistence
-silent! setlocal formatoptions+=j
 
 " smartindent will be overridden by indentexpr if filetype indent is on, but
 " otherwise it's better than nothing.


da-x:
I have `formatoptions=croqnlj` and I cannot reproduce said behavior. 
Any other reason why it breaks in your setup?

Thomasdezeeuw:
I think you need `a` in you initial options.

da-x:
`a` in 'initial options'? I am not sure what you mean.

Thomasdezeeuw:
You need to set the `a` options in the format options, i.e. `formatoptions=croqnlja`, before loading the rust plugins. I think if you add after it might also work (`:set formatoptions+=a`).

da-x:
Ok, got it.

With `formatoptions=croqnlja` the problem still does not reproduce. The comment's paragraph is reformatted property when adding text, and when hitting enter, the comment breaks up properly with respect to `//`.

```
// Some awesome long
// comment
```

Thomasdezeeuw:
It seems a faulty `formatlistpat` setting was to blame here, which I happen to fix along side with changing the `formatoptions`. Sorry for the noise.

PR #330, opened 2019-06-08
Create highlight group for string delimiters

bk2204:
Some users wish to highlight string delimiters (e.g., quotation marks) in a different color from the body of the string itself. This feature is provided by several other languages' syntax highlighting rules, and is easy to achieve for Rust as well.

Add a match group for the string delimiters and by default, assign it to the same highlight as the rest of the string, to preserve existing behavior. Users who wish to make use of this feature need only adjust their color scheme files to link the new `rustStringDelimiter` group to their preferred syntax highlighting group.


bk2204:
Gentle ping.  Is it possible that someone could review this when it's convenient?

PR #331, opened 2019-06-29
Highlight delimiters

arzg:
Previously delimiters such as `{` and `}` were left unhighlighted.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -137,6 +137,7 @@ syn match     rustOperator     display "&&\|||"
 " so it skips the ->; see http://stackoverflow.com/a/30309949 for details.
 syn match     rustArrowCharacter display "->"
 syn match     rustQuestionMark display "?\([a-zA-Z]\+\)\@!"
+syn match     rustDelimiter    "[(){}\[\];]"
 
 syn match     rustMacro       '\w\(\w\)*!' contains=rustAssert,rustPanic
 syn match     rustMacro       '#\w\(\w\)*' contains=rustAssert,rustPanic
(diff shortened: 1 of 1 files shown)


arzg:
I’ve also added a commit that highlights `<` and `>` (e.g. in type params) as Delimiters instead of the Operators they’re currently highlighted as.

chris-morgan:
I oppose this change: this isn’t what `Delimiter` is for. As `:help group-name` says:

> ```
> 	 Delimiter	character that needs attention
> ```

Braces and brackets don’t need attention.

A few Vim syntax files use `Delimiter` for braces or brackets (its name *does* lend itself towards such implementation, and some people are generally very trigger-happy about highlighting things specially), but it’s a small minority.

Given my opposition and the lack of any remark from anyone else since this was filed, I’m going to close this. Further arguments in its favour may be made, but if you desire it I suggest you just keep it in your `~/.vim/after/syntax/rust.vim`.

arzg:
Hm, interesting – didn't realise that's what Delimiter was for. I guess it's quite ironic in that case that quite a number of Vim colourschemes I've used _de-emphasise_ Delimiter. Is there an appropriate highlight group for what I'm trying to accomplish with my use of Delimiter here?

chris-morgan:
You call it de-emphasising, but I’d tend to say that simply being *different* provides a certain degree of emphasis.

Anyway: no, there isn’t. Vim historically hasn’t highlighted those things specially; I’d say that highlighting them differently is a comparatively recent idea, in the scheme of things.

arzg:
Sorry, what I meant with de-emphasising was reducing their contrast, similar to what many colourschemes do with comments.

> Anyway: no, there isn't

Well, that's a shame. I've seen `Noise` being used by some JavaScript syntax plugins for exactly this purpose, but not much elsewhere. Since linking/highlighting `Noise` is to be taken care of by the colourscheme (and is sadly often left untouched), it would still make sense IMHO to use that instead of `Delimiter`.

Issue #313, opened 2019-03-09
Error with tagbar addon: E716: Key not present in Dictionary: n


reinerh:
When using rust.vim with tagbar, vim throws an error in Rust files that contain a module when opening the tagbar.
(When using also the `airline` addon, it shows the error immediately on start, as airline is also using tag data for the status line.)
Without rust.vim (i.e., only with the tagbar addon active), the tagbar is shown correctly and without any errors (that's why I'm filing this bug here and not for tagbar).

* rust.vim version: fdb2d39 (Feb 18 2019)

Steps to reproduce:

Create a new library cargo project.
```
$ cargo new --lib hello
```

Open src/lib.rs with vim and open tagbar with `:Tagbar`.

Expected behavior:

vim displays the tagbar with tags of the current .rs file.

Actual behavior:

vim throws this error:

```
Error detected while processing function tagbar#ToggleWindow[2]..<SNR>37_ToggleWindow[9]..<SNR>37_OpenWindow[83]..<SNR>37_AutoUpdate[63]..<SNR>37_ProcessFile[84]..<SNR>37_ParseTagline[76]..<SNR>37_ProcessTag[74]..<SNR>37_add_tag_recursive[
96]..<SNR>37_create_pseudotag[32]..<SNR>43_initFoldState:
line   13:
E716: Key not present in Dictionary: n
```

RustInfo:

```
rust.vim Global Variables:

let g:ftplugin_rust_source_path = v:null
let g:loaded_syntastic_rust_cargo_checker = v:null
let g:loaded_syntastic_rust_filetype = v:null
let g:loaded_syntastic_rust_rustc_checker = v:null
let g:rust_bang_comment_leader = v:null
let g:rust_cargo_avoid_whole_workspace = v:null
let g:rust_clip_command = v:null
let g:rust_conceal = v:null
let g:rust_conceal_mod_path = v:null
let g:rust_conceal_pub = v:null
let g:rust_fold = v:null
let g:rust_last_args = v:null
let g:rust_last_rustc_args = v:null
let g:rust_original_delimitMate_excluded_regions = v:null
let g:rust_playpen_url = v:null
let g:rust_prev_delimitMate_quotes = v:null
let g:rust_recent_nearest_cargo_tol = v:null
let g:rust_recent_root_cargo_toml = v:null
let g:rust_recommended_style = v:null
let g:rust_set_conceallevel = v:null
let g:rust_set_conceallevel=1 = v:null
[...]

PR #333, opened 2019-07-18
use default tagbar settings with universal ctags

deathlyfrantic:
I'm not sure this is the best way to do this, but using the settings in this plugin will break Tagbar with Universal Ctags (see #313). [Tagbar already has more complete ctags definitions](https://github.com/majutsushi/tagbar/blob/master/autoload/tagbar/types/uctags.vim#L784-L820) for Universal Ctags than the settings here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,9 @@ for more details.
 
 ### Source browsing with [Tagbar][tgbr]
 
-`rust.vim` has builtin `ctags/` definitions for [Tagbar][tgbr] which
-are loaded if Tagbar is installed.
+The installation of Tagbar along with [Universal Ctags](uctags) is recommended
+for a good Tagbar experience. For other kinds of setups, `rust.vim` tries to
+configure Tagbar to some degree.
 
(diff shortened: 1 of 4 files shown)


da-x:
Hi, thanks for the PR. I mostly just need to allocate some time to verify that this is the correct solution, before merging it as-as. Worth checking whether other language plugins are doing something better.

da-x:
Thanks for the extensive research!

Surely, we should not override Tagbar if it's already fully capable of supporting Rust, given the Universal Ctags is installed.

Sadly, there doesn't seem to be an interface for Tagbar to tell us if it has detected Universal Ctags. Actually, we almost had one - `tagbar#inspect("ctags_is_uctags")` works only if Tagbar was already used once in the session, and force-initing does not seem to be good solution. So your change seems best course of action.

To improve it, I pushed to your branch. Added caching of the ctags version detection. Tagbar caches it as well. Also, extended the documentation. Please tell whether it works for you and we can merge this.

PR #334, opened 2019-07-23
shell escape path

Konfekt:


chris-morgan:
I didn’t recall this functionality, so I looked it up: it was added in [Vim 7.4.191](https://github.com/vim/vim/commit/26df092843de91ea0c5c5c130d0d0695d2d81c07). We still *kinda* support versions that old. I would therefore prefer this to be guarded with `has("patch-7.4.191")`. (This `has()` form requires patch 7.4.237, but I’m OK with this fix not applying for 7.4.191–7.4.236, that’s a rare enough range that `v:version > 704 || (v:version == 704 && has("patch191"))` seems not worth it.)

Konfekt:
It now checks for the patch version. Is escaping `%` is necessary? In case the patch is not included, then the file path is expanded and shell escaped. Thus there is no perfect solution: if the file is renamed, `&makeprg` uses a non-existent file path. If the file path is not escaped, then possibly as well.

compiler/rustc.vim
```diff
@@ -21,7 +21,11 @@ endif
 if get(g:, 'rustc_makeprg_no_percent', 0)
     CompilerSet makeprg=rustc
 else
-    CompilerSet makeprg=rustc\ \%
+    if has('patch-7.4.191')
+      CompilerSet makeprg=rustc\ \%:S
+    else
+      exe 'CompilerSet makeprg=rustc ' . shellescape(expand('%'))
```
chris-morgan: I meant just doing without this, `CompilerSet makeprg=rustc\ \%`. You can’t trust `%` to remain the same over time (e.g. start with `:new | setf rust` and only give it a filename later).

Konfekt: Exactly, 

> In case the patch is not included, then the file path is expanded and shell escaped. Thus there is no perfect solution: if the file is renamed, &makeprg uses a non-existent file path. If the file path is not escaped, then possibly as well.

There is no perfect solution.

chris-morgan: Ah yes, I didn’t read very thoroughly, did I? But yeah, I’d prefer it to just skip the escaping. Simpler (which is important for legacy code that isn’t the normal code path), typically still right.

Konfekt: Okay. Let's hope for the best! Coders rarely use file paths with anything but alphanumeric letters. Different for markup languages.

chris-morgan:
Good commit message. ☺

Issue #307, opened 2019-02-09
Incorrect indentation for closing parenthesis of tuple literal


drmikehenry:
# Version information

* rust.vim version: Jan 10, 2019 (the latest as of this writing)

* Vim version: 8.1.1453

# Steps to reproduce

* Create an empty environment with only rust.vim:

```sh
mv ~/.vimrc ~/orig.vimrc
mv ~/.vim ~/orig.vim
git clone https://github.com/rust-lang/rust.vim ~/.vim
```
  
* Create `test.rs` containing the below source, formatted as a multi-line [tuple literal](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/expressions.md#tuple-literals) with a final comma:

```rust
fn main() {
    let some_tuple = (
        "first long string",
        "second long string",
        "third long string",
        "fourth long string",
    );
}
```

* Re-indent the entire file via ``gg=G``.

# Expected vs. actual behavior

The initial indentation in the file is correct, as verified via `rustfmt --check test.rs`.  After indentation using Vim, the tuple's closing parenthesis becomes shifted to the right to align with the final string:

```rust
fn main() {
    let some_tuple = (
        "first long string",
        "second long string",
        "third long string",
        "fourth long string",
        );
}
```

`rustfmt --check test.rs` shows this indentation to be incorrect:

```diff
Diff in /home/drmikehenry/test.rs at line 4:
         "second long string",
         "third long string",
         "fourth long string",
-        );
+    );
 }
``` 

Note that the indentation is correct if the final comma is removed from the last string in the tuple, but leaving the comma out is incorrect as verified by `rustfmt --check test.rs`.

# Debugging info from RustInfoToFile
```
rust.vim Global Variables:

let g:ftplugin_rust_source_path = v:null
let g:loaded_syntastic_rust_cargo_checker = v:null
let g:loaded_syntastic_rust_filetype = v:null
let g:loaded_syntastic_rust_rustc_checker = v:null
let g:rust_bang_comment_leader = v:null
let g:rust_cargo_avoid_whole_workspace = v:null
let g:rust_clip_command = v:null
let g:rust_conceal = v:null
let g:rust_conceal_mod_path = v:null
let g:rust_conceal_pub = v:null
let g:rust_fold = v:null
let g:rust_last_args = v:null
let b:rust_last_args = []
let g:rust_last_rustc_args = v:null
let b:rust_last_rustc_args = []
[...]

PR #335, opened 2019-07-25
Incorrect indentation for closing parenthesis of tuple literal

ANtlord:
https://github.com/rust-lang/rust.vim/issues/307

### DIFF
--- a/indent/rust.vim
+++ b/indent/rust.vim
@@ -11,8 +11,8 @@ endif
 let b:did_indent = 1
 
 setlocal cindent
-setlocal cinoptions=L0,(0,Ws,J1,j1,m1
-setlocal cinkeys=0{,0},!^F,o,O,0[,0]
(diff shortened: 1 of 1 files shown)


indent/rust.vim
```diff
@@ -11,16 +11,16 @@ endif
 let b:did_indent = 1
 
 setlocal cindent
-setlocal cinoptions=L0,(0,Ws,J1,j1,m1
-setlocal cinkeys=0{,0},!^F,o,O,0[,0]
+setlocal cinoptions=L0,(0,Ws,J1,j1,m1,(s,m1
```
chris-morgan: Replace the `(0` with `(s`, rather than adding it on the end. I agree that `(0` should go, that’s an artefact of when visual indentation was preferred in some places; it is no longer used anywhere in the standard formatting guide.

And `m1` is already there.

```suggestion
setlocal cinoptions=L0,(s,Ws,J1,j1,m1
```

PR #337, opened 2019-08-12
Fix joining comment line removes // comment leader

### LINKED ISSUE #78: Line joining with J removes `//` from comment
Line joining with `J` on line 1 here:

```rust
1:  foo();
2:  // comment
```

Produces incorrect result:

```rust
1: foo() comment
```

### LINKED ISSUE #336: Joining comment line removes // comment leader
I found https://github.com/rust-lang/rust.vim/issues/78 as the same issue, but it was closed. So I created a new one. This occurs only on Vim 7.3.541 or later since `formatoptions` supports `j`.

I have already located the root cause of this issue. So I will also create a PR soon.

* rust.vim version: 772eaa6a991d2a0f53844f241a6af3bc1ab1fc50

## Steps to reproduce:

Open below buffer in Vim with `ft=rust`.

```rust
fn main() {
    println!("hello");
    // hello
}
```

Then type `J` at `println!(...)` line.

### Expected

Joining the comment line as follows

```rust
fn main() {
    println!("hello"); // hello
}
```

### Actual

`//` is omitted

```rust
fn main() {
    println!("hello"); hello
}
```

Paste debugging info from the Rust Vim plugin via _one_ of the following
commands: `:RustInfo`,  `:RustInfoToClipboard`, or `:RustInfoToFile <filename>`.

<details>
<pre>
rust.vim Global Variables:
let g:rust_conceal = v:null
let g:rust_fold = v:null
let g:rust_last_args = v:null
let b:rust_last_args = []
let g:rust_recommended_style = 0
let g:rustc_path = v:null
let g:rustfmt_autosave = 0
let g:rustfmt_emit_files = 1
let g:rustfmt_options = ''
let g:syntastic_extra_filetypes = ['rust', 'crystal']
rustfmt 1.2.2-stable (5274b49c 2019-04-24)
rustc 1.36.0 (a53f9df32 2019-07-03)
cargo 1.36.0 (c4fcfb725 2019-05-15)

VIM - Vi IMproved 8.1 (2018 May 18, compiled Jul 21 2019 22:54:00)
macOS version
Included patches: 1-1722
Compiled by Homebrew
Huge version with MacVim GUI.  Features included (+) or not (-):
+acl               +dnd               +listcmds          +postscript        +textobjects
+arabic            -ebcdic            +localmap          +printer           +textprop
</pre>
</details>

rhysd:
Fixes #336, #78 

The root cause of #336 and #78 is that the line joined with comment line is wrongly regarded as comment line. Vim uses `comments` option to check if the line is comment or not. And in `ftplugin/rust.vim`, it is set to include

```
m:\ ,
```

It means that a line starts with one space is middle of block comment. So Vim wrongly regards any line which starts with space as comment line.

```
    A
    // B
```

When joining a comment line `// B` into `A` with `J`, Vim regards `A` line is a comment line. So it removes `//` from `// B` to join two comment lines.

I don't know why `m:\ ` was included in `comments` option, but it looks some mistake for me. So this patch removes the config and after that, `J` will work fine fixing #336 and #78.

### DIFF
--- a/ftplugin/rust.vim
+++ b/ftplugin/rust.vim
@@ -35,7 +35,7 @@ if get(g:, 'rust_bang_comment_leader', 0)
     " leaders. I'm fairly sure that's a Vim bug.
     setlocal comments=s1:/*,mb:*,ex:*/,s0:/*,mb:\ ,ex:*/,:///,://!,://
 else
-    setlocal comments=s0:/*!,m:\ ,ex:*/,s1:/*,mb:*,ex:*/,:///,://!,://
+    setlocal comments=s0:/*!,ex:*/,s1:/*,mb:*,ex:*/,:///,://!,://
 endif
 setlocal commentstring=//%s
 setlocal formatoptions-=t formatoptions+=croqnl


chris-morgan:
This particular bug has annoyed me mildly from time to time, but never enough to investigate.

Apparently I introduced this in July 2013 with 650496e5c5fe5e5c92c4c368e4454274e79edf28.

I observe that the current behaviour of `/*` comments doesn’t match what I wrote there anyway, insofar as it *does* add a ` * ` leader when you’re just typing, and if you remove it prefers to leave a two-space leader. I am disinclined to worry about this myself, however, because the convention these days is to use line comments exclusively.

I suspect that this lack of conformance with what I documented may be significant to the presence of `m:\ `, as well as the comment from a few lines above:

```vim
    " Why is the `,s0:/*,mb:\ ,ex:*/` there, you ask? I don't understand why,
    " but without it, */ gets indented one space even if there were no
    " leaders. I'm fairly sure that's a Vim bug.
```

Sure, it’s not exactly the same, but it’s plausibly linked. But like you I’m not really sure. I’m quite happy to shrug and move on.

I can’t find any problems with the result of this change, having tried `//` and `//!`, and each of `/*`, `/**` and `/*!` with and without `*` leaders on inner lines; all at top level and indented; with `J` with the four combinations of the current and next lines being a comment and a non-comment line, and with `gqG`. The only change I observe is the desired one, that `J` on a line with content followed by a `//`, `//!` or `///` line doesn’t remove the comment leader any more.

rhysd:
Thank for confirming that. Since I'm not fully understanding the side effect by removing the config from the option, please feel free revert this patch if it causes some issue.

arcnmx:
From what I can tell, not using `m:\ ` removes the middle leader of `/*!` comments and results in a lot of weirdness when using those comments? For example, `i/*! hello<cr>hi<esc>Othere` results in:
```rust
/*! hello
/*! there
*/ hi

chris-morgan:
Evidently I was doing something wrong when I tested things. I now call into suspicion my entire results. 😖

PR #338, opened 2019-08-15
Correct check for rustfmt_autosave_if_config_present

DoumanAsh:
It still doesn't help me to prevent rustfmt from running though so I just turned all auto runs on save

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -239,7 +239,7 @@ function! rustfmt#PreWrite()
     if !filereadable(expand("%@"))
         return
     endif
-    if rust#GetConfigVar('rustfmt_autosave_if_config_present', 0)
+    if rust#GetConfigVar('rustfmt_autosave_if_config_present', 1)
         if findfile('rustfmt.toml', '.;') !=# '' || findfile('.rustfmt.toml', '.;') !=# ''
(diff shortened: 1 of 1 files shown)


da-x:
I don't think this is the correct fix. Did you reload Vim or the buffers after changing `g:rustfmt_autosave_if_config_present`? If not, there's an alternative fix.

DoumanAsh:
I'm not sure I follow the code very well.

Butt currently my .vimrc has:

```
let g:rustfmt_autosave = 0
let g:rustfmt_autosave_if_config_present = 0
let g:rustfmt_autosave_because_of_config = 0
```

And it was running rustfmt anyway. 
And yes I did even restart vim.

Right now I just removed rustfmt code completely because I don't use it, but as of now I have no idea where and why it runs rustfmt

DoumanAsh:
Re-checking it once, again, it does indeed makes no difference.
So for now I just remove this rustfmt completely and use local fork

da-x:
I tested with these three config lines without reproducing the issue.

BTW the `let g:rustfmt_autosave_because_of_config = 0` line should not be there, as it's not really a global variable to be changed. There is indeed a problem if `g:rustfmt_autosave_if_config_present` is  changed without reloading Vim and I'm fixing that.

PR #341, opened 2019-08-21
fix 'RustTest' fail when vim build with feature `-terminal`

pendulm:
error screen:
image

when VIM build with feature `terminal` off, the command `:terminal` is still available but report error. So, check the feature with `has()` instead `exists()`

:version report
image

:terminal output
```
:terminal
E319: Sorry, the command is not available in this version

:echo has('terminal')
0

:echo exists(':terminal')
2
```

### DIFF
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -505,7 +505,7 @@ function! rust#Test(all, options) abort
         return rust#Run(1, '--test ' . a:options)
     endif
 
-    if exists(':terminal')
+    if has('terminal')
         let cmd = 'terminal '
     else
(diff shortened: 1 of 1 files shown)


JeanMertz:
This unfortunately breaks terminal support on neovim. I get the same output as shown by @pendulm, but in my case, terminal support is actually enabled (it's always enabled in neovim.

```
:echo has('terminal')
0

:echo exists(':terminal')
2
```

I guess the proper solution would be to do

```vim
if has('terminal') || has('nvim')
```

I'll make a PR.

Issue #345, opened 2019-08-28
Unknown option: cfg


zenixls2:
* rust.vim version:  git commit f45cdae27133912e6c1c1d41c17b9613b60afe07
Steps to reproduce:

mine installation is through vim-plug.  
by changing the order of these two lines, different results will occur:  
```vim
Plug 'rust-lang/rust.vim'
Plug 'cespare/vim-toml'
```

Expected vs. actual behavior:

if vim-toml is loaded before rust.vim, no error will occur.
otherwise, vim will output something like:
```text
"Cargo.toml" 10L, 252C
Error detected while processing FileType Auto commands for "*":
E518: Unknown option: cfg
```
and vim-toml will be ignored (no color for the script).  
The error message is produced by [this line](https://github.com/rust-lang/rust.vim/blob/master/ftdetect/rust.vim#L4)
vim version is 
```text
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Sep 18 2016 14:42:40)
```

Paste debugging info from the Rust Vim plugin via _one_ of the following
commands: `:RustInfo`,  `:RustInfoToClipboard`, or `:RustInfoToFile <filename>`.

these won't work.

mhinz:
You didn't mention the `Included patches` bit from `vim --version`, but you probably have a version older than [`8.0.0613`](https://github.com/vim/vim/commit/3e54569b1), which is the commit that introduced the `FALLBACK` syntax.

As for the order issue: When `cespare/vim-toml` is run first, it [sets the filetype to `toml`](https://github.com/cespare/vim-toml/blob/master/ftdetect/toml.vim#L2) and thus the [`:setfiletype` from this plugin](https://github.com/rust-lang/rust.vim/blob/master/ftdetect/rust.vim#L4) will return early without ever processing its arguments.

zenixls2:
the version I used is 8.0.003-1 provided in Ghettoforge Plus repo in CentOS 7.
the default verson in main repo is at 7.4

Is there any possibility to do the version check before to prevent this issue?
Or maybe we need to add a dependency requirement of version 8.0.0613 in README.

mhinz:
The above PR should fix it for you.

PR #346, opened 2019-08-28
`:setfiletype FALLBACK` is not available in older versions

mhinz:
There is no workaround for this. If we would set the filetype first via `:set filetype=cfg` or `:setfiletype cfg`, more specialized plugins like [vim-toml](https://github.com/cespare/vim-toml), wouldn't be able to overwrite the filetype with `:setfiletype toml` anymore.

Now, we simply don't set a filetype at all for older Vim versions. Not great, but better than an error that keeps rust.vim from getting sourced properly.

Fixes https://github.com/rust-lang/rust.vim/issues/345

---

Alternative: We remove the entire line. If someone wants syntax highlighting for toml, [vim-toml](https://github.com/cespare/vim-toml) should be used.

p.s. Works with Vim and Nvim.
p.p.s This is the recommended approach for checking versions (in older Vim verions): `:h has-patch`.

### DIFF
--- a/ftdetect/rust.vim
+++ b/ftdetect/rust.vim
@@ -1,7 +1,10 @@
 " vint: -ProhibitAutocmdWithNoGroup
 
 autocmd BufRead,BufNewFile *.rs call s:set_rust_filetype()
-autocmd BufRead,BufNewFile Cargo.toml setf FALLBACK cfg
+
+if has('patch-8.0.613')
+    autocmd BufRead,BufNewFile Cargo.toml setf FALLBACK cfg
+endif
 
 function! s:set_rust_filetype() abort
     if &filetype !=# 'rust'


ftdetect/rust.vim
```diff
@@ -1,7 +1,10 @@
 " vint: -ProhibitAutocmdWithNoGroup
 
 autocmd BufRead,BufNewFile *.rs call s:set_rust_filetype()
-autocmd BufRead,BufNewFile Cargo.toml setf FALLBACK cfg
+
+if v:version > 800 || v:version == 800 && has('patch613')
```
chris-morgan: You don’t need to use the old technique; `if has('patch-8.0.613')` will work correctly.

mhinz: :+1: I actually checked the docs, but apparently didn't read far enough: https://github.com/vim/vim/blob/v8.0.0613/runtime/doc/eval.txt#L8406-L8413

PR #351, opened 2019-09-19
Highlight label after break and continue statements as rustLabel

### LINKED ISSUE #159: Loop annotations are not highlighted correctly.
On current master, loop annotations are highlighted one way where they are defined, and a complete different way when they are used.

Here is on my MacVim 8.0.329 with theme colors chosen to be as close as possible to Monokai from Sublime Text.

image

Here's the same code in Atom with colors from the monokai community theme (you can tell what theme I like).

image

rhysd:
Fixes #159 

I manually confirmed this fixes #159 and does not break current syntax highlighting for lifetime parameters. Here is a screenshot:

スクリーンショット 2019-09-19 13 15 45

Only `'label`s after `break` and `continue` statements are highlighted as `rustLabel`

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -205,6 +205,7 @@ syn region rustGenericLifetimeCandidate display start=/\%(<\|,\s*\)\@<='/ end=/[
 "rustLifetime must appear before rustCharacter, or chars will get the lifetime highlighting
 syn match     rustLifetime    display "\'\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*"
 syn match     rustLabel       display "\'\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*:"
+syn match     rustLabel       display "\%(\<\%(break\|continue\)\s*\)\@<=\'\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*"
 syn match   rustCharacterInvalid   display contained /b\?'\zs[\n\r\t']\ze'/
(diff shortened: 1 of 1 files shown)


syntax/rust.vim
```diff
@@ -205,6 +205,7 @@ syn region rustGenericLifetimeCandidate display start=/\%(<\|,\s*\)\@<='/ end=/[
 "rustLifetime must appear before rustCharacter, or chars will get the lifetime highlighting
 syn match     rustLifetime    display "\'\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*"
 syn match     rustLabel       display "\'\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*:"
+syn match     rustLabel       display "\%(\<\%(break\|continue\)\s\+\)\@<=\'\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*"
```
chris-morgan:
```suggestion
syn match     rustLabel       display "\%(\<\%(break\|continue\)\s*\)\@<=\'\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*"
```
Whitespace isn’t required between `break` and the label. This lets you write code like this if you want to:
```rust
fn main() {
    let over = ();
    's: loop {
        break's over
    }
}
```

rhysd: I did not know that. Thanks. I'll update this patch.

rhysd:
I updated this patch to consider no space before label in `break` and `continue` statements.

スクリーンショット 2019-09-19 14 57 14

syntax/rust.vim
rhysd: I addressed this at 4928319

PR #352, opened 2019-09-20
Highlight `in` with the same color as `for`

rhysd:
Current rust.vim highlights `for` with `rustRepeat` but `in` with `rustKeyword`.

As far as I checked libsyntax in rustc implementation, `in` keyword is currently used only for `for` statement. I want to highlight `in` with the same color as `for` to indicate it is a part of `for` loop statement.

Here is a screenshot before/after applying this patch.

- Before

スクリーンショット 2019-09-20 13 00 32

- After

スクリーンショット 2019-09-20 13 00 52


da-x:
Thanks!

PR #356, opened 2019-10-05
Refactoring for s:RmDir() in autoload/rust.vim

rhysd:
- Check the directory exists using `isdirectory()`. If it does not exist, we don't need to remove it
- Check given path is not home and root after ensuring the path is expanded. Relative path does not match to them
- Use `delete()` built-in function rather than executing `rm` command. It works on Windows and we don't need to take care about path escaping.


da-x:
Thanks!

PR #357, opened 2019-10-08
Fix: link in README

matthiasbeyer:
This patch fixes a link in the README file.


da-x:
Fixed by other means - needed square brackets.

PR #359, opened 2019-10-14
Use separate buffer in Neovim

lambdalisue:
While the behavior of 'terminal' in Vim and Neovim is different, all Neovim users cannot get back to the original buffer if the terminal command has directly applied.
To solve the issue and make the behavior consistent, execute 'new' prior to 'terminal' in Neovim.

## Without this PR
https://asciinema.org/a/42wz3Z9LuKlylLTeZkk7rhuUU

## With this PR
https://asciinema.org/a/BFkMEcdYCKn3EnWwDXixcckoB

### DIFF
--- a/autoload/rust.vim
+++ b/autoload/rust.vim
@@ -514,8 +514,10 @@ function! rust#Test(all, options) abort
         return rust#Run(1, '--test ' . a:options)
     endif
 
-    if has('terminal') || has('nvim')
+    if has('terminal')
         let cmd = 'terminal '
+    elseif has('nvim')
+        let cmd = 'new | terminal '
     else
         let cmd = '!'
         let manifest = shellescape(manifest)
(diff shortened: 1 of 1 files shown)


autoload/rust.vim
```diff
@@ -514,8 +514,10 @@ function! rust#Test(all, options) abort
         return rust#Run(1, '--test ' . a:options)
     endif
 
-    if has('terminal') || has('nvim')
+    if has('terminal')
         let cmd = 'terminal '
+    elseif has('nvim')
+        let cmd = 'new | terminal '
```
mhinz:
```suggestion
        let cmd = 'noautocmd new | terminal '
```

Many plugins run their code on `BufEnter`, `WinEnter` etc. Avoid this using `:noautocmd`.

lambdalisue: Nice thanks.

rhysd:
Nice PR :+1:. I imlemented `:RustTest` but I did not consider Neovim very well at that time.

PR #361, opened 2019-10-27
Use :terminal for running cargo

rhysd:
`:RustTest` uses terminal feature but cargo commands don't. This PR makes cargo commands such as `Cbuild`, `Ctest`, ... use `:terminal` to run `cargo` command.

tmp


da-x:
Thanks!

PR #371, opened 2019-11-22
Highlight the unit type

arzg:
Inspired by [the recent post in r/rust](https://www.reddit.com/r/rust/comments/dwa1jm/unit_highlighting_in_emacs/), I have added unit type highlighting. The regex was definitely quick and dirty, but it seems to work for me in most cases.


chris-morgan:
This plugin has a history of following the most common Vim convention, of not highlighting parentheses, braces and brackets.

And there’s nothing special about unit—it’s just an empty tuple.

On these grounds, I oppose this change. Highlighting of tuples in general, that could be a thing, satisfying the second objection if not the first—but I’d still expect to oppose it because I’d expect it to have a substantial performance cost: with a regular expression parser it’s fairly computationally expensive to select only the parentheses that are of tuples.

If you really want this, it is fortunately easy for you to put in your after/syntax/rust.vim.

Issue #249, opened 2018-08-14
Highlight all types from Rust Standard Library


zachwhaley:
* rust.vim version: rust.vim version built with NeoVim v0.2.2 "Last Changed" date in rust.vim file is Feb 24, 2016

Common Standard types like Vec and String are highlighted, yet less common types like VecDeque and OsString are not.

chris-morgan:
Only prelude types are highlighted. Adding other types from the standard library but *not* highlighting other types from *outside* the standard library is not useful—there’s nothing special about types in libstd beyond the prelude that would warrant special highlighting for them. Highlighting all types is the subject of #142. Closing in favour of that.

Issue #142, opened 2017-03-01
Suggestion: syntax/rust.vim: Add highlighting for user-defined types


Ploppz:
At the moment, I have this extra line (maybe not perfect) in `rust.vim`:
`syn match Type "\v[A-Z]\w*[a-z0-9]"`

Furthermore, if it's decided to add something like this, it would be nice if it could have precedence over `rustModPath`, for cases like `Object::new();`.

chris-morgan:
Pretty sure there was an issue about this somewhere (maybe back in the days when this was in the rust-lang/rust repository?). My own preference with it was not to do this. Personally I think the capital letter is enough of a signal, and excessive highlighting is icky in most color schemes. As it is, we just highlight the prelude items.

zacps:
Why don't we make it optional then? I'd personally like this added and I doubt I'm alone.

chris-morgan:
If we do this, I think it’d make sense to have three “modes”:

1. Don’t highlight any items specially;
2. Highlight prelude items specially (as we currently do); and
3. Highlight CamelCase words as types, and maybe do something about tuple struct and tuple variant syntax (c.f. `rustEnumVariant` that we have at present).

I would suggest still keeping the current behaviour as the default, in order to (a) not change things gratuitously (well, I would perceive it as gratuitous), and (b) not mishighlight things when working with code bases that don’t adhere to the rustfmt conventions (e.g. working with the Windows API). Yet still this might be my own biases to simple highlighting showing.

PR #372, opened 2019-11-25
Add some more built-in types

arzg:
I know this isn’t a full list, but I thought it might be nice to have anyway.


chris-morgan:
We’ve discussed this a couple of times before (e.g. #249); there’s really nothing special about non-prelude std types, and even them I’m not completely sold on highlighting these days (and I think I was the one that added them). I believe that adding more types individually would be a mistake, and I’m declining this change for that reason.

Might be time to think more about #142, the more general approach to this.

PR #373, opened 2019-11-25
Highlight variable definitions and constant names

lunacookies:
The regexes I’ve used are definitely not perfect, I just quickly slapped them together. Also, constants that begin with a number are not currently highlighted.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -64,6 +64,8 @@ syn match   rustExternCrateString /".*"\_s*as/ contained nextgroup=rustIdentifie
 syn keyword   rustObsoleteExternMod mod contained nextgroup=rustIdentifier skipwhite skipempty
 
 syn match     rustIdentifier  contains=rustIdentifierPrime "\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*" display contained
+syn match     rustIdentifier  "\v((let|static|const) (ref )?(mut )?)@<=[a-z_]+"
+syn match     rustIdentifier  "\v<[A-Z_][A-Z0-9_]+>" " Leave out the numbers on the first char to avoid matching numbers
 syn match     rustFuncName    "\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*" display contained
 
 syn region rustMacroRepeat matchgroup=rustMacroRepeatDelimiters start="$(" end=")" contains=TOP nextgroup=rustMacroRepeatCount


chris-morgan:
```vim
syn match     rustIdentifier  "\v((let|static|const) (ref )?(mut )?)@<=[a-z_]+"
```

This is insufficient: e.g. `let Foo(bar) = baz;`, these things are patterns, not identifiers. Also there are three other constructs that can introduce new bindings (match branches, function arguments, closure arguments). In general, I don’t like things that make it work *sometimes* but not *always*. I don’t think a *good* solution exists, when using a regular expression parser. You *could* make it highlight identifiers inside a pattern in each of these cases, but it’d slow things down a lot, and I don’t think that people actually care so very much about the precise highlighting, but they do about the performance of the highlighter. (Do take what I say with a grain of salt, however: I’m biased towards simple highlighting, and have `Identifier` cleared.)

> constants that begin with a number are not currently highlighted.
I’m not sure what you’re meaning—if you mean “the form of identifier that is conventionally used for constants and statics”, then there’s no such thing as one beginning with a number.

syntax/rust.vim
```diff
@@ -64,6 +64,8 @@ syn match   rustExternCrateString /".*"\_s*as/ contained nextgroup=rustIdentifie
 syn keyword   rustObsoleteExternMod mod contained nextgroup=rustIdentifier skipwhite skipempty
 
 syn match     rustIdentifier  contains=rustIdentifierPrime "\%([^[:cntrl:][:space:][:punct:][:digit:]]\|_\)\%([^[:cntrl:][:punct:][:space:]]\|_\)*" display contained
+syn match     rustIdentifier  "\v((let|static|const) (ref )?(mut )?)@<=[a-z_]+"
```
chris-morgan: This is insufficient: e.g. `let Foo(bar) = baz;`, these things are patterns, not identifiers. Also there are three other constructs that can introduce new bindings (match branches, function arguments, closure arguments). In general, I don’t like things that make it work *sometimes* but not *always*. I don’t think a *good* solution exists, when using a regular expression parser. You *could* make it highlight identifiers inside a pattern in each of these cases, but it’d slow things down a lot, and I don’t think that people actually care so very much about the precise highlighting, but they do about the performance of the highlighter. (Do take what I say with a grain of salt, however: I’m biased towards simple highlighting, and have `Identifier` cleared.)

chris-morgan (commented):
> constants that begin with a number are not currently highlighted.

I’m not sure what you’re meaning—if you mean “the form of identifier that is conventionally used for constants and statics”, then there’s no such thing as one beginning with a number.

arzg:
> you mean “the form of identifier that is conventionally used for constants and statics”

Yep, correct.

> then there’s no such thing as one beginning with a number

Cool, I wasn’t sure exactly where to look for the answer to something like this.

syntax/rust.vim
arzg: This might be a good start for tuple destructuring:

    syn match rustIdentifier "\v((let|static|const) (ref )?(mut )?)@<=([a-z_]+|\((\w+(, )?)+\))"

It currently matches the parentheses and commas as well though.

The following works for simple cases (i.e. no destructuring) with `for` loops:

    syn match rustIdentifier "\v(for (&)?)@<=\w+"

PR #374, opened 2019-11-25
Match all enum variants

arzg:
This PR uses the fact that functions follow snake_case to avoid matching CamelCase enum variants as functions. A possible downside is that some may not follow these conventions, but I get the impression that they are near-universal in the Rust community.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -101,9 +101,7 @@ syn keyword rustTrait Default
 syn keyword rustTrait Iterator Extend IntoIterator
 syn keyword rustTrait DoubleEndedIterator ExactSizeIterator
 syn keyword rustEnum Option
-syn keyword rustEnumVariant Some None
 syn keyword rustEnum Result
-syn keyword rustEnumVariant Ok Err
 syn keyword rustTrait SliceConcatExt
 syn keyword rustTrait String ToString
 syn keyword rustTrait Vec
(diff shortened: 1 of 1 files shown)


chris-morgan:
Bear in mind that this syntax is for tuple structs (`struct Foo(…);`) as well as tuple variants.

When fully within the Rust ecosystem, yes, such conventions are fairly strong. But once you get into foreign interfaces, e.g. Windows development, it can be very common to have camel case function names. (Sometimes things will be renamed to match Rust’s conventions, either in situ or through a safe interface atop the low-level bindings, other times not.)

Despite that uncertainty, I think I’d be content (just barely) to consider camel case followed by a parenthesis to be either a tuple struct/variant instantiation, or destructuring.

But as a matter of fact, I generally oppose disambiguation between tuple struct/variant instantiation and function calls—the former *is* a function call (when in expression position, anyway; function calls aren’t a thing in patterns).

I think that it’s just barely reasonable to special-case `Some`, `None`, `Ok` and `Err` as prelude items and as ones used almost universally—that we were highlighting prelude types was the whole reason why `rustEnumVariant` became a thing at all.

Anything further, though, is getting into #142 territory, where *perhaps* it might make sense to offer three “modes”: no special highlighting, prelude types only (as currently) and full guessing (where types and tuple structs/variants will be highlighted differently). At present I think I’m going to close this (as with #372) in favour of a more considered approach to satisfying #142.

Issue #105, opened 2016-09-13
rust.vim should respect the user's indentation style.


mortie:
Currently, rust.vim changes indentation to be four spaces. Personally, I strongly prefer tabs (though please don't make this into a tabs vs spaces argument). I think rust.vim should respect the user's choice, not forcefully change indentation style to four spaces.

chris-morgan:
Vim defaults to tabs. The standard Rust ecosystem is strongly four-spaces,
so rust.vim really needs to default to four spaces. Simply setting the
options as we do now is really just about the best way to do it.
after/syntax/rust.vim is a suitable place to override it—or if there isn't
a 'g:rust_something' option for our we could add such a thing. (I don't
have the code handy right now.)

I did put quite a lot of thought into these sorts of things some time back;
I tend to agree with you that forcing things like that isn't good, while
saying that defaulting to the standard conventions is good. It's a bit of a
balancing act.

dengelt:
> The standard Rust ecosystem is strongly four-spaces, so rust.vim really needs to default to four spaces. 

I don't understand why this makes it necessary to override the user's settings.

ExpHP:
Despite my *infinite, undying hatred of soft-tabs in every. single. editor. that has supported them ever*—I am going to defend CM's position in that I do not believe an unreasonable choice has not been made here.  For those individuals who don't mind either way, it is a good thing to push them towards the community standards.

**As for motivated individuals such as us who *do very much* mind, we can easily find a workaround!**

This plugin is actually a bit kinder than some, because it actually comes with [a "please don't do that" option](https://github.com/rust-lang/rust.vim/blob/fc11d02fee330df7b30c83a80f09dd0c60ab43ce/ftplugin/rust.vim#L41-L44)!  Setting `let g:rust_recommended_style = 0` in vimrc lets you keep your tabs.

Of course, there's also a bunch of other things which it will do regardless of those settings, like changing `formatoptions`.  For times like these, I usually find that [venting one's frustration into a .vim/after/ file](https://github.com/ExpHP/dotfiles/blob/master/vim/after/ftplugin/python.vim) is quite a successful strategy.  In this case, the place where you want to put your overrides is `~/.vim/after/ftplugin/rust.vim`.

steveklabnik:
Yes, you can set that option if you don't want to use the recommended style.

PR #379, opened 2020-01-02
Don't set formatoptions

george-b:
These are a matter of user preference.

### DIFF
--- a/ftplugin/rust.vim
+++ b/ftplugin/rust.vim
@@ -37,9 +37,6 @@ else
     setlocal comments=s0:/*!,ex:*/,s1:/*,mb:*,ex:*/,:///,://!,://
 endif
 setlocal commentstring=//%s
-setlocal formatoptions-=t formatoptions+=croqnl
-" j was only added in 7.3.541, so stop complaints about its nonexistence
-silent! setlocal formatoptions+=j
 
 " smartindent will be overridden by indentexpr if filetype indent is on, but
 " otherwise it's better than nothing.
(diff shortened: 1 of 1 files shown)


da-x:
Meddling with `formatoptions` is done by the majority (63%) of other filetypes bundled with Vim. Please provide more information in support of this?

george-b:
Yup, whilst lots of them do this I don't think it's something that should be perpetuated. I [raised this in the main Vim repo](https://github.com/vim/vim/issues/5406) and Bram suggested I contact individual maintainers. This is my first report of such so I'm very much testing the waters here.

da-x:
Per the discussion in #105, looked more into this, and I think `rust.vim` provides a better default than Vim for this option. If you want to override it, please use an `after/ftplugin/rust.vim` file. I see that Bram also hinted at that.

george-b:
I agree with filetypes touching indentation and tab settings as those are pertinent.

My view is this that the behaviours `formatoptions` defines for certain interactions is a user preference. I'd imagine one would want to be consistent across all filetypes, without having to add configuration.

For what it's worth I have the following in my config to ensure such consistency for all filetypes. But I certainly consider it a hack.

```
augroup FormatOptions
  autocmd!
  autocmd FileType * let &l:formatoptions = &g:formatoptions
augroup END
``` 

If this removing the setting of `formatoptions` isn't of interest feel free to close this.

da-x:
We'll keep it then. It also may be one of those things is too late to remove without causing a bit of disruption.

PR #382, opened 2020-01-12
use poor man's shellescape in case %:S unavailable

Konfekt:
Might be worth a try. Found this, e.g., in `$VIMRUNTIME/compiler/jikes.vim`.


da-x:
Merged. We have greater chance of having spaces in filenames than having `"`.

Issue #340, opened 2019-08-20
:Ccheck not an editor command


GopherJ:

mhinz:
That is correct. There is [no such command](https://github.com/rust-lang/rust.vim/blob/master/plugin/cargo.vim#L8-L21).

rhysd:
I think you can use instead

```
:Cargo check
```

da-x:
Fixed by #383.

PR #383, opened 2020-01-19
Add :Ccheck

yankuu:


yankuu:
The Rust Edition Guide [recommended](https://doc.rust-lang.org/edition-guide/rust-2018/cargo-and-crates-io/cargo-check-for-faster-checking.html) `cargo check` for faster checking. This is also mentioned in #340.

da-x:
Thanks!

PR #384, opened 2020-02-03
Add g:cargo_shell_command_runner to specify how to run cargo commands

### LINKED ISSUE #376: Regression?: Cargo commands open in a new buffer
I recently went back to a setup that I haven't used since maybe February or March, and I naturally had to update everything. When I did, I noticed that rust.vim's various cargo commands, like `Cbuild`, `Cargo`, etc. now output to a separate buffer, as opposed to just outputting like normal `!` commands do.

This might not be *so* bad if it re-used the buffer, though it would still be a pain on small screens, but it does not. Every single invocation opens a new buffer, which means that every buffer must be manually closed. It's easier for me just to write "!cargo build".

I can't see an option in the docs to turn this off, so I personally consider it a serious usability regression.

rhysd:
Fixes #376

This PR adds `g:cargo_shell_command_runner` to customize how to run shell commands for cargo.

For example, by setting `"!"` to it, `:!` command is used to run `cargo ...` commands on `:Cargo`, `:Cbuild`, `:Crun`, ...

```vim
let g:cargo_shell_command_runner = '!'
```

### DIFF
--- a/autoload/cargo.vim
+++ b/autoload/cargo.vim
@@ -1,12 +1,14 @@
-function! cargo#Load() 
+function! cargo#Load()
     " Utility call to get this script loaded, for debugging
 endfunction
 
-function! cargo#cmd(args)
+function! cargo#cmd(args) abort
     " Trim trailing spaces. This is necessary since :terminal command parses
     " trailing spaces as an empty argument.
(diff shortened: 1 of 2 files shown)


da-x:
Thanks!

PR #387, opened 2020-02-17
Allow bar to be placed after :RustFmt

PurpleMyst:
This PR allows constructs such as:

```vim
autocmd BufWritePre <buffer> if &modified | RustFmt | endif
```

Which currently fail with `E488`.

The "feature" is added by specifiying `-bar` when defining `RustFmt`


da-x:
Thanks!

Issue #262, opened 2018-09-10
plugin doesn't seem to work?


MarkSwanson:
* rust.vim version:  github master, enabled via pathogen

:set filetype?
filetype=rust

:set runtimepath?
runtimepath=~/.vim,~/.vim/bundle/YouCompleteMe,~/.vim/bundle/ale,~/.vim/bundle/html5.vim,~/.vim/bundle/poppy.vim,~/.vim/bundle/rust.vim,~/.vim/bundle/vim-airline,~/.vim/bundle/vim-airline-themes,~/.vim/bundle/vim-easyclip,~/.vim/bundle/vim-jav
ascript,~/.vim/bundle/vim-repeat,/var/lib/vim/addons,/usr/share/vim/vimfiles,/usr/share/vim/vim81,/usr/share/vim/vimfiles/after,/var/lib/vim/addons/after,~/.vim/bundle/rust.vim/after,~/.vim/bundle/vim-javascript/after,~/.vim/after

vim 8.1 - patch 209
rustc 1.30.0-nightly (f7202e40f 2018-08-27)

Expected vs. actual behavior:

:RustFmt
expected: format stuff
actual: Not an editor command: RustFmt

Paste debugging info from the Rust Vim plugin via _one_ of the following
commands: `:RustInfo`,  `:RustInfoToClipboard`, or `:RustInfoToFile <filename>`.

:RustInfo

== Not an editor command: RustInfo

*EDIT* :Cargo build works fine, so the plugin seems to be at least partially working?

da-x:
Most likely that Vim is not loading the last version, but perhaps its builtin older version instead.
Are the Rust files files having syntax highlighting enabled?

MarkSwanson:
Yes, syntax highlighting works.
I renamed/appended .bak to all of my installed vim81/*.*rust*vim files.
That didn't seem to help. I also backed up my vim81 dir and deleted all the *rust* files, but that didn't help.

MarkSwanson:
Also note: I installed rustfmt via rustup. It's in ~/.cargo/bin/rustfmt
$ rustfmt --version
rustfmt 0.99.2-nightly (5c9a2b6 2018-08-07)

Maybe "-nightly" is confusing the version check?

da-x:
I have rustfmt 0.99.4-nightly installed and it works. 

Maybe you can bisect your `rust.vim` back to a version that worked, to see what broke it?

MarkSwanson:
[...]

PR #389, opened 2020-02-22
Update Readme to correctly load the `RustFmt` command

robinmoussu:
When using vim8 installation method, `RustFmt` isn't a valid command.
After some researches, I found
https://github.com/rust-lang/rust.vim/issues/262 which suggested to add
`filetype indent on` in my .vimrc witch solved the issue.

Note: I'm not sure if this should be added in the `RustFmt` section or the
`vim8` section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ configure Tagbar to some degree.
 ### Formatting with [rustfmt][rfmt]
 
 The `:RustFmt` command will format your code with
-[rustfmt][rfmt] if installed.
+[rustfmt][rfmt] if installed and if you have `filetype plugin on`
+in your `.vimrc`.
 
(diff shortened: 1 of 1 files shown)


da-x:
Which plugin manager are you using?

robinmoussu:
vim8 built-in functions

everybodyeverybody:
Hi,

I was about to open a PR to update this documentation this as well.

My only comment is I think this should be in the installation section of the README, rather than down where the command for formatting is specified, as it is necessary to get that functionality to work and thus should be done on installation.

da-x:
@everybodyeverybody I'm considering to merge #412 considering the various issues similar to this.

da-x:
#412 merged.

PR #390, opened 2020-02-26
Fix highlighting raw identifier

rhysd:
I noticed that [raw identifiers](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/raw-identifiers.html) are not highlighted correctly.

スクリーンショット 2020-02-27 1 46 59

Note: `r` in `fn r#match` is highlighted as `rustFuncName` but `#match` is highlighted as `rustMacroName`. They happen to have the same color.

This PR fixes the highlight as follows:

スクリーンショット 2020-02-27 1 47 38


da-x:
Thanks!

PR #400, opened 2020-04-20
Avoid polluting jumplist while running :RustTest

rhysd:
I noticed that jumplist was unexpectedly updated while running `:RustTest`. It was updated because it searched the end of test function with `%`. This PR makes the command not to change jumplist.


da-x:
Thanks!

PR #409, opened 2020-05-31
Add configurable g:rust_cargo_check_cmd for syntax checking

zzmp:
Adds configuration to change the cargo command used for syntax checking.
Without this, using a custom toolchain will fail to find the core crate.


da-x:
This is good, but it should use `rust#GetConfigVar` so it can be set per buffer via mechanisms such as [vim-localvimrc](https://github.com/embear/vim-localvimrc) or [local_vimrc](https://github.com/LucHermitte/local_vimrc/).

PR #410, opened 2020-06-10
Syntax for asm! macro (RFC 2850)

ayosec:
This patch adds support for the syntax added in https://github.com/rust-lang/rust/pull/69171 for the `asm!` macro.

As an example, these screenshots show the difference in the highlight of the macro.

* [Current master](https://user-images.githubusercontent.com/37484/84216921-58e0fc00-aaba-11ea-9f60-adcd813751a7.png)
* [After these changes](https://user-images.githubusercontent.com/37484/84216918-57173880-aaba-11ea-914f-00a2504b10dc.png)


da-x:
Thanks!

PR #415, opened 2020-07-08
Accept modifiers in expand/emit commands

bobrippling:
This enables commands like `:vertical RustEmitAsm` to split the new window vertically, and so on.


bobrippling:
Bump?

PR #418, opened 2020-09-10
Highlight SAFETY comments

Flying-Toast:


Flying-Toast:
I'm not sure if emacs and VSCode also highlight TODO/FIXME/... comments like this. If they do we should also update them.

chris-morgan:
I haven’t come across `// SAFETY: …` comments before, but I see there are about 615 occurrences of this pattern in the standard library (`rg SAFETY | wc -l` in ~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library), and it makes sense as a style for Rust, so it seems reasonable to include. Thanks!

PR #419, opened 2020-09-16
Use winsaveview()/winrestview() instead of :mkview/:loadview for restoring the window view

itchyny:
The current implementation fails to restore the cursor position when

- it fails to store the view due to incorrectly configured 'viewdir'
- 'viewoptions' does not include 'cursor'

Using winsaveview()/winrestview() can restore the view with no side-effect.

### DIFF
--- a/autoload/rustfmt.vim
+++ b/autoload/rustfmt.vim
@@ -107,7 +107,7 @@ function! s:DeleteLines(start, end) abort
 endfunction
 
 function! s:RunRustfmt(command, tmpname, from_writepre)
-    mkview!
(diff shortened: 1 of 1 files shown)


da-x:
Thanks! Originally the plugin did use these functions, but I changed it without being aware of the side effects. BTW the doc for `winsaveview()` does say "This may have side effects.". I wonder what those are...

itchyny:
Oops, I didn't noticed that in the doc. By side effect, I mean :mkview saves the view in the local file while winsaveview() does not. Not sure what the doc means though. Thanks for merging.

PR #420, opened 2020-09-21
Adds support to :RustTest command for non standard test libraries

jkogara:
The current implementation of `SearchTestFunctionNameUnderCursor` does not detect `[tokio::test]` as being a valid test attribute, this change adds a wildcard to the start of the test detection regexp.


Pranav2612000:
Are there any plans to merge this in? Happy to help any way I can?

Issue #414, opened 2020-07-04
:RustTest doesn't work on async tests with macros


GopherJ:
* rust.vim version:  latest

Steps to reproduce:
1. put cursor on an async test
![image](https://user-images.githubusercontent.com/33961674/86511552-9fe0b980-bdfa-11ea-8952-f542c5264b46.png)
2. execute :RustTest

Expected vs. actual behavior:

Paste debugging info from the Rust Vim plugin via _one_ of the following
commands: `:RustInfo`,  `:RustInfoToClipboard`, or `:RustInfoToFile <filename>`.

_?_

davidarmstronglewis:
I've run into maybe a similar issue with the error: ```error[E0670]: `async fn` is not permitted in the 2015 edition``` which makes me think that I haven't properly told `rust.vim` something about which edition of Rust I'm using (my Cargo.toml specifies 2018 edition). Is this the error you're seeing or something different?

GopherJ:
@davidarmstronglewis I got this:

![image](https://user-images.githubusercontent.com/33961674/88489945-215dde80-cf98-11ea-95c1-e0ec392493d4.png)

I think it's not the same thing

PR #423, opened 2020-10-12
Added support for async tests (and other libraries)

spadarian:
At the moment the regex used to find tests are a bit limmited. They don't recognise non standard libraries (#420) or async functions (#414).

I think this PR fixes both.

Maybe there are other prefixes for `fn` (I just added async), so feel free to suggest others.


rogercoll:
Any update on this? I think it would be very helpful to have the same benefits of "RustTest" in asynchronous tests

PR #430, opened 2020-12-10
Struct Definition Improvements

### LINKED ISSUE #429: Highlight struct names in `rustFoldBraces`
Hello! Thank you guys for your work on Rust– let alone this project.

There is one small oddity with the highlighting I've noticed, however: mainly, that `struct` names are not highlighted when used as a `struct` field. These are typically located inside the `rustFoldBraces` region.

For example:

```rust
struct Foo
{
    foo: Bar
}
struct Bar {…}
```

The `Bar` in `foo: Bar` should be highlighted the same as `Foo` in `struct Foo`. Right now `rustIdentifier` is used for `Foo` but not for `Bar` .

It seems that primitives are detected though:

```rust
struct Bar
{
    bar: u8
}
```

Iron-E:
Closes #429.

I thought I should try to close the issue I created— some of the decisions I made might seem questionable but I'm happy to either explain them further or work with others until it's in a place that this is worth merging.

This PR does a few things of direct relation to #429:

1. Introduce a new `rustStructDefinition` group specifically for structures.
    * Add `rustFieldName` for struct field names.
    * Add `rustFieldSep` for separators of field names from field types.
2. Add `rustNoise` to allow highlighting of commas, semicolons, and dot operators.
3. Add `rustConstraint` to allow highlighting of `:` in `where Foo : Bar` from a struct definition.
4. Remove unutilized generics region implementation in favor of `rustGeneric`.
5. Add syntax highlighting for curly braces using `matchgroup`.

Topics left to discussion:

* [ ] Default highlight links for new groups?
    * I have some defaults created [here](https://github.com/Iron-E/nvim-highlite/blob/master/colors/highlite.vim#L581-#L591) but I'm not insistent on imposing them as _the_ defaults.

### DIFF
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -14,6 +14,8 @@ endif
 
 " Syntax definitions {{{1
 " Basic keywords {{{2
+syn match     rustNoise "[,.;]"
+syn match     rustConstraint "\s\+:\s\+"
 syn keyword   rustConditional match if else
 syn keyword   rustRepeat loop while
 " `:syn match` must be used to prioritize highlighting `for` keyword.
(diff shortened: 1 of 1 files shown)


chris-morgan:
This is a fairly significant departure from the stylistic conventions of Vim in general and this plugin specifically, and it’s necessarily very incomplete, per the notes I just put in #429. I’m sorry, but I don’t think it fits into rust.vim, though you’re welcome to try making a deliberately more full-featured syntax file. (But really I’m more interested in the idea of shifting *all* syntax highlighting to rust-analyzer, which Vim 8.2 made theoretically possible.)

Iron-E:
I have to say I tend to prefer LSP highlighting to regex highlighting.

That said, I don't really think this "deviates" from Vim's convention. Having worked extensively with many other language syntax files, I can definitely say that the naming and structure are derivative of other projects. Additionally, I did look at other groups to choose appropriate names within the project.

This is one of the few places in a project where we know where types are going to be, and can highlight them accordingly. Other areas I definitely understand the reluctance.

Here is before:



Here is after:



I'll just maintain the fork and direct my users to it until either full support for user types is developed here or highlighting is moved to LSP.

Thanks for looking it over!

PR #431, opened 2020-12-11
User Type Highlighting

### LINKED ISSUE #21: Inconsistent highlighting of `super` in imports
In imports like `use super::super::math::{Vec3, Mat3};` the second (and possible 3rd, 4th, ...) `super` is highlighted like a keyword although the first one is highlighted like a module name.

### LINKED ISSUE #102: Feature request: Highlight user-defined types
One thing I miss is for user-defined types to be assigned a highlight group, so that I can give them the same color as basic types such as `i32`, `usize`, `String`, etc.
By user-defined types, I mean types that you as a programmer have defined in your project/crate.
Eventually also for types imported from other crates.

If there are performance issues, maybe the plugin should update its table/index/whatever of keywords on explicit command.

### LINKED ISSUE #142: Suggestion: syntax/rust.vim: Add highlighting for user-defined types

### LINKED ISSUE #205: Incorrectly highlighting tuple structs as functions
I've noticed that tuple structs are often being highlighted as functions, which leads to some confusion because the highlighting appears inconsistent. Notably, in an `enum` body:

```rust
pub enum Message {
  Greeting(String), // highlighted as function - notice no space before paren
  Goodbye (String), // highlighted as normal text - notice space before paren
  Wave, // highlighted as normal text
  NamedGreeting { // highlighted as normal text
    target: String,
    words: String,
  }
}
```

You can see that even GH has an issue with inconsistency here - the empty option (`Wave`) and the struct option (`NamedGreeting`) are both highlighted as normal text, while the other two are highlighted as a function call. `rust.vim` only highlights the `Greeting` option. Because different items of the `enum` are highlighted differently, it can lead to some confusion because the highlighting doesn't actually mean anything in those cases - it's accidental.

I realize there's some difficulty determining whether an identifier is a function call, a tuple struct, etc, but I think there's a few solutions. My first idea would be to highlight *all* `enum` members, as well as the final identifier in scoped lists (eg, this same problem occurs with `phrasing::Message::Greeting(...)` vs `phrasing::Message::Wave` - the `Greeting` is highlighted while the `Wave` isn't, leading to a false sense of difference between the two).

Thanks for reading!

### LINKED ISSUE #397: Syntax highlighting incorrectly reports char literals as lifetimes
* rust.vim version: 255a81f

Steps to reproduce:

Try to use a char such as `'_'` or `'c'`

Expected vs. actual behavior:

omg_rust
**Expected**: the highlight group should be character (`rsCharacter`)
**Actual**: the highlight group is marked as lifetime (`rsInferredLifetime` and `rsUserLifetime`)

```
rust.vim Global Variables:

let g:rust_conceal = v:null
let g:rust_fold = v:null
let g:rust_last_args = v:null
let b:rust_last_args = []
let g:rustc_path = v:null
let g:rustfmt_autosave = 1

let g:rustfmt_emit_files = 1
let g:rustfmt_options = ''
rustfmt 1.4.11-stable (9eb4b564 2020-01-29)
rustc 1.42.0 (b8cedc004 2020-03-09)
cargo 1.42.0 (86334295e 2020-01-31)

NVIM v0.4.3
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/local/Homebrew/Library/Homebrew/shims/mac/super/clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedanti
c -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagn
ostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/tmp/neovim-20191107-85913-1wbgqz6/neov
im-0.4.3/build/config -I/tmp/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/src -I/usr/local/include -I/tmp/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/deps-build/i
nclude -I/usr/local/opt/gettext/include -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include -I/tm
p/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/build/src/nvim/auto -I/tmp/neovim-20191107-85913-1wbgqz6/neovim-0.4.3/build/include
Compiled by brew@Catalina.local

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/usr/local/Cellar/neovim/0.4.3/share/nvim"

```

Iron-E:
Closes #21.
Closes #102.
Closes #142.
Closes #205.
Closes #373.
Closes #397.
Closes #428.
Closes #477.

___

Yesterday I opened a PR and it was closed. Despite this, the reviewer said:

> you’re welcome to try making a deliberately more full-featured syntax file.

So I did. Now, types are highlighted in most places rather than just struct definitions. I used active community repositories as a benchmark, such as Ripgrep.

Further, if changes to my chosen names are desired, just _ask_ :). I'm sure we can figure something out.

This PR does a few things of direct relation to #142:

1. Add `rustNoise` to allow highlighting of commas, semicolons, dot operators, and other noise characters.
2. Add `rustBounds` to allow highlighting of `:` in `where Foo : Bar` from a struct definition.
3. Remove unutilized generics region implementation in favor of `rustGenericRegion`.
4. Add syntax highlighting for fold regions using `matchgroup`.
5. Add syntax highlighting for all user-defined types through `@rustIdentifiers`.
6. Rename types to fit expected naming from other syntax files in Vim community, according to the table below:
	| Old Name         | New Name          |
	|:-----------------|:------------------|
	| `rustIdentifier` | `rustType`        |
	| `rustType`       | `rustBuiltinType` |
	| \<variables>     | `rustIdentifier`  |
	| \<constants>     | `rustConstant`    |
	* See `:h group-name` for more information.

Topics left to discussion:

* [x] Default highlight links for new groups?
	* I have some defaults created [here](https://github.com/Iron-E/nvim-highlite/blob/master/colors/highlite.vim#L581-#L591) but I'm not insistent on imposing them as _the_ defaults.

___

If you find some place where there wasn't coverage let me know.

<table>
	<tr>
		<th><code>master</code></th>
		<th>This PR</th>
	</tr>
	<tr>
		<td>master</td>
		<td>PR</td>
	</tr>
</table>


Iron-E:
Currently working to relax the detection of `rustType` because it's detecting too much.

Iron-E:
I will clean up the commit history if desired, but I usually leave it in case its useful for anyone else to look at.

Up to the reviewer.

da-x:
Thanks @Iron-E, this is amazing work! Largest update to the syntax description in quite awhile. @chris-morgan please help review this?

Iron-E:
Sorry for that last round of commits— there were some more issues that I uncovered which needed to be ironed out. Haven't discovered anything else in about a week since though so it's probably close in terms of function.

I will say that there are some things which this implementation doesn't try to do, like tell the difference between typedefs, traits, and the like. There really isn't a great way to know for sure whether or not something is just a "type"/struct or if it's something more specific without leveraging a language server (which I believe is also being discussed).

So, these changes would probably act as a fallback for when the LSP is not available (or as a bridge until the LSP highlighting is integrated).

Iron-E:
Found some more fixes. A lot of this was already here in master, but the changes in the PR revealed it. I figured I should fix the issues before any merging of the changes was done

lidin:
Hi, I noticed identifiers containing numbers are not highlighted correctly. I believe the following changes remedy that:

```
diff --git a/syntax/rust.vim b/syntax/rust.vim
index fe926fc..1df2399 100644
--- a/syntax/rust.vim
+++ b/syntax/rust.vim
@@ -56,13 +56,13 @@ syn keyword   rustAwait       await nextgroup=@rustIdentifiers skipempty skipwhi
 syn match     rustKeyword     /\<try\>!\@!/ display


-syn match  rustIdentifier "\v<\l+(_+\l+)*>" contained display
-syn match  rustConstant   "\v<\u+(_+\u+)*>" contained display
+syn match  rustIdentifier "\v<\l(\l|\d)*(_+(\l|\d)+)*>" contained display
+syn match  rustConstant   "\v<\u(\u|\d)*(_+(\u|\d)+)*>" contained display

 syn match  rustFuncName   "\v<\w+>((::)?\<.*\>)?\s*(\()@=" contains=rustModPathSep,rustGenericRegion display
 syn region rustAnonymousFunc matchgroup=rustFuncName start="|" end="|" contains=rustNoise,rustBounds,rustStorage,@rustIdentifiers

-syn match  rustType       "\v<\u>|<\u+\l+(\u+\l*)*>" contains=rustEnum,rustEnumVariant,rustTrait,rustDeriveTrait nextgroup=rustModPathSep display
+syn match  rustType       "\v<\u>|<\u+(\l|\d)+(\u+(\l|\d)*)*>" contains=rustEnum,rustEnumVariant,rustTrait,rustDeriveTrait nextgroup=rustModPathSep display

 syn match  rustUnused "\v<_" contained contains=rustIdentifier display

@@ -116,8 +116,8 @@ syn cluster rustScopes contains=rustSuperScope,rustSelfScope,rustCrateScope

 syn region rustPubScope matchgroup=rustNoise start='(' end=')' contained contains=rustRepeat,@rustScopes,rustModPath,rustModPathSep transparent

-syn match   rustModule      "\v<\l+(_+\l+)*>" contained contains=@rustScopes display
-syn match   rustModPath     "\v<\l+(_+\l+)*>(\s*::(\<)@!)@=" contains=rustModule display nextgroup=rustModPathSep
+syn match   rustModule      "\v<\l(\l|\d)*(_+(\l|\d)+)*>" contained contains=@rustScopes display
+syn match   rustModPath     "\v<\l(\l|\d)*(_+(\l|\d)+)*>(\s*::(\<)@!)@=" contains=rustModule display nextgroup=rustModPathSep
 syn match   rustModPath     "\v(^\s*(pub\s+)?(use|mod)\s+)@<=<\w+>(\s*::<\w+>)*;@=" contains=rustModule,rustType display
 syn match   rustModPathSep  "::" nextgroup=rustModPath,@rustIdentifiers display skipempty skipwhite
```

Iron-E:
@lidin I already have that fixed over on [my master branch](https://github.com/Iron-E). Thanks for noticing it here though!

There are so many commits on this PR that it got to a point that I didn't want to inundate the maintainers all at once with all my changes. If you want to use the most up-do-date version of this "edition" of `rust.vim` you should check out my repo, that's where I've been pushing all my fixes recently.

lidin:
Well, I would think it is always better to keep the changes in one place, but it is ultimately up to the maintainers to decide whether this change should be merged or not. Personally, I don't think the changes are that crazy or that many to not be merged, but there is perhaps only the aspect of how it affects performance that would need to be taken into account. I am not very proficient in the "Vim syntax" syntax, so I'm not sure what affects performance or how to test it properly.

Iron-E:
I personally don't believe that my changes affect performance all that much but I know that there has been [some concern](https://github.com/rust-lang/rust.vim/pull/373#discussion_r350027004) that changes to the identifier highlighting could. I am running on an old 2010 Macbook, and it never stutters or lags out. I don't think it will be a problem but everyone gets an opinion on that ;)

I was waiting to get this resolved before starting other activities, but ended up just getting to work on `Iron-E/master` after a few months of indecision on that front. Since my divergence I've moved some stuff around in the file to make things closer to each other's logical groupings, changed the highlight defaults to follow the convention in `:h group-name`, and fixed quite a few issues with the current highlight implementation. 

I'm glad you want these features! Maybe I can decouple the bug fixes from the default highlighting and make the default highlighting a separate PR (since that is definitely out of scope), if the default highlighting revamp is desired

@da-x any thoughts?

da-x:
There seems to be an oddity with the names of the first struct fields, for example:

image

Second one has `rustIdentifier`, the first one doesn't.

Iron-E:
Ah. I think I know why that is happening: the `,`, which registers as `rustNoise` (open to renaming that) has a `nextgroup` which makes the `rustIdentifier` be detected. I just have to add logic to detect the `rustIdentifier` first.

I didn't notice because I have `Identifier` linked with `Normal` on my colorscheme. Thanks for catching this!

I'll experiment with various fixes tomorrow and push the cleanest one I come up with

Iron-E:
I pushed the changes. I tested it on a bunch of random files and situations just to see if the fix worked without breaking highlighting anywhere else, and I believe this is sound. Let me know if you see anything off though

da-x:
Is it really pushed? I don't see the new commit.

BTW, please add a 'demo rs' file to the repo that tests out various aspects of the syntax? I am considering to integrate https://github.com/da-x/vim-syntax-legend into the testing.

Iron-E:
Sorry for the confusion. I've been pushing to my master branch since I was not sure the scope of changes you were open to accepting (see [this comment](https://github.com/rust-lang/rust.vim/pull/431#issuecomment-795435557)). Would you like me to rebase this PR onto my master branch? It has more fixes than this PR but larger scope.

As for a demo file, I was just thinking about that  yesterday. Too cumbersome to run around checking files in different repos. My commit messages have some edge cases listed so I can get started on that.

da-x:
Let's extend the scope of the PR - there's a lot of very useful work in there I'd like myself and subscribing parties give it a go for awhile. If we find bugs after merging we can fix them separately.

da-x:
Also, to raise to your attention, after testing your master branch - there seems to be an issue with `rustFoldBraces` matching.

I use the function below [1] to probe syn matching. The amount of `rustFoldBraces` you should be seeing should relate to the nesting level, IIUC. For example:

```rust
use ksite_backend::database::{self, DatabaseExecutor, GitState};
use kverse::ibash;
```

First `use` gives: `['rustKeyword, trans: Keyword']`.
However, second `use` gives: `['rustFoldBraces, trans: rustFoldBraces', 'rustKeyword, trans: Keyword']`.

In a long `.rs` the probing over `fn main` gives a very long list of `rustFoldBraces`, but it doesn't make sense as we are in the top fold level.

[1]
```vim
" Show highlight names at cursor

function! IdentifySyn(...) abort
  let l:stack = []
  for l:i in synstack(line('.'), col('.'))
    call add(l:stack, synIDattr(l:i, 'name') . ', trans: ' . synIDattr(synIDtrans(l:i), 'name'))
  endfor
  echo l:stack
endfunction

command! IdentifySyn call IdentifySyn()
```

Iron-E:
> Let's extend the scope of the PR - there's a lot of very useful work in there I'd like myself and subscribing parties give it a go for awhile. If we find bugs after merging we can fix them separately.

Sounds good! I was sort of hoping you'd be of that opinion. I will integrate the changes from the `master` branch over to here.

> In a long .rs the probing over `fn main` gives a very long list of `rustFoldBraces`, but it doesn't make sense as we are in the top fold level.

That is strange indeed… I'll investigate this. I'll try that function you posted as well. Up until now I had been using this:

```vim
command! EchoHlGroup echo "Highlight Group: " . synIDattr(
\	synID(line("."), col("."), 1), "name"
\) | echo 'Namespace: ' . synIDattr(
\	synID(line("."), col("."), 0), "name"
\) | echo "Linked to: " . synIDattr(
\	synIDtrans(synID(
\		line("."), col("."), 1
\	)), "name"
\)
```

Iron-E:
I've added a bunch of stuff to the demo file. There isn't any one place which has long strings of syntax like you might expect in a real Rust file, but there's lots of different little tests for different variations.

Let me know if there's any other syntax I haven't tested. There probably is— Rust is a big language.

da-x:
@Iron-E I've been experimenting with the branch to see that it looks good under various popular colorschemes. One thing I noticed is the `Identifer` being rather pervasive, not letting the neutral color in. Maybe my configuration is wrong? But it look this is a property of the branch itself.

Take for example https://github.com/nanotech/jellybeans.vim , I see:

image

Now, for what I see both screenshot demoing the colorscheme and casual browsing of Rust code on Github usually have some neutral color like white being given for what are usually identifiers. I'm aware we you can set or clear `rustIdentifer` or `Identifer` in vimrc to address this, but I think this may surprise users getting this update.

Also compare gruvbox with its syntax highlighting gallery: https://github.com/morhetz/gruvbox/wiki/Gallery .

image

da-x:
Also, found that `unsafe` matching works in standalone statements but not in assignments:

image

Iron-E:
We can `hi def link rustIdentifier Normal`. It might be strange for some who expect `Identifier` highlighting but I honestly don't think most would question it.

I'll fix `= unsafe`, and I'll add a test for it in the demo. Thanks for catching that!

da-x:
Thanks for keeping this up. I've noticed that `path` (lowercase) gets `Type` matching everywhere, is this intended?

Iron-E:
Not intended. I'll see if I can get the macro-specific types to only highlight after `:`

Iron-E:
No rush, just letting you know in case you missed it that I fixed that issue from before. If you've found any other issues with the PR feel free to let me know 👍

da-x:
Working with this for a week, I did not find further issues.

However since it's a major change I'd like to expand testing to more users before this hits the multitude of users' environments.

Maybe @joelpalmer, @seandewar, @tversteeg, @lemunozm would like to test?

joelpalmer:
Hi, I've mostly been using language-agnostic and some Rust-specific treesitter groups for a while now.   

But, TS has been giving me some trouble over the last few days. So, I've tested most of the groups in this PR and things look good. Perfect timing! 

I'll stay on this branch and hammer on it today. I'll report back at my EOD _iif_ I run into issues. Otherwise, it LGTM.

Iron-E:
Just a heads-up, I went through a the open issues and added which ones this PR should close to the top.

Dispersia:
thank you for this! While a lot of it is a massive improvement, my keyword 'struct' is now remapped to the color of variables, while before it was the same color as 'fn'
image

phi-gamma:
Thanks a lot for your work, this already looks great!

However I’ve got some trouble with enums. The ``enum`` keyword
now is colored the same as the type name and variants:

Now

Whereas before I got the keyword distinct from the enum type name,
the enum type name receiving the same color as variants, and types
like ``u8``, ``bool`` (builtins only?) in another color:

Before

Using a tweaked 16-color *solarized* palette in the terminal with the
vanilla *solarized* color scheme.

Iron-E:
@Dispersia thank you for providing your config! I looked at your colorscheme, and it turns out that onedark doesn't differentiate between the `Structure` highlight group and `Type`. This will probably be common— I'll change the default highlighting of `rustStructure`/etc to `Keyword` for clarity 👍

___

@phi-gamma thanks for trying this out! The `enum` keyword issue should be fixed once I push the above. 

As for the reason that `Color` was highlighted blue before: it used to just be an "`Identifier`" before this PR, which is blue in [solarized](https://github.com/altercation/vim-colors-solarized/blob/master/colors/solarized.vim#L552) (and interesting fact, Variants were `Function` for a long time even on this PR). Now they are highlighted as `Type` along with the built-ins. You can still change this by setting `rustBuiltinType` or `rustType` to be different.

As for why constants look like literals: constants were not recognized on master, so you are probably seeing `Statement` there. UPPER_SNAKE_CASE tokens are now shown as `Constant`. Since [solarized](https://github.com/altercation/vim-colors-solarized/blob/master/colors/solarized.vim#L544) highlights constants and literals the same way, it shows no difference. You can highlight `rustConstant` as `Statement` again if desired. 

I think I see where solarized is coming from though, since a constant is similar to a literal.

phi-gamma:
> @phi-gamma thanks for trying this out! The `enum` keyword issue
> should be fixed once I push the above. 

Fantastic, thank you!

>                You can still change this by setting
> `rustBuiltinType` or `rustType` to be different.

I’m going to do that then, thanks for digging into this!

> I think I see where solarized is coming from though, since a
> constant is similar to a literal.

Indeed!

-phg

kevinclark:
Hi there! Just wanted to chime in. I'm a [solarized8](https://github.com/lifepillar/vim-solarized8/blob/master/colors/solarized8.vim) user and the default experience feels somehow too colorful? Variables start to blend in and the rustNoise tokens get color.

Before:

Screen Shot 2021-11-22 at 9 31 32 PM

After:

Screen Shot 2021-11-22 at 9 31 14 PM

Being able to distinguish between categories seems like a great thing. Any suggestions for how to clean up my colorscheme for the brave new world you're pushing? I'm happy to RTFM if you can point me in the right direction.

Dotfiles [here](https://github.com/kevinclark/dotfiles/tree/master/dotfiles/config/nvim) if you're curious. They're up to date sans a line to plug in this branch.

Iron-E:
> Any suggestions for how to clean up my colorscheme for the brave new world you're pushing?

I'd love to!

For context: there are a set of conventions for Vim syntax highlights in `:h group-name`. This is what I based the PR off of.

Unfortunately, a good number of colorschemes don't use these conventions. Some don't know about them, others disagree with them, certain schemes only use 16 colors and thus need to reuse in certain areas. This is where the problem lies 😅

For example, I link `rustNoise` to `Delimiter` (since they are delimiters). However solarized links that group with `Special`, which is why there is so much visual clutter. Another example is for `Identifier`— solarized uses a deep blue for that, so now when all of the `Identifier`s are detected as intended it looks very colorful. Ideally common / less-important groups have less color 

There are a couple things you can do:

1. Keep doing as you've done, and manually re-link highlight groups that you want to change (since they can now be targeted). 
    * The more colors you have available to you, the easier this is. If you only have 16 colors available, you will probably have to reuse colors in certain areas. `cterm` for italics/bold can also help to differenciate.
    * As I understand it, the thing is to make important stuff stand out and give everything else a more neutral tone.
2. [Look for a 256-color scheme](https://vimcolorschemes.com/dark/) which also follows the conventions. These ones look pretty good:
    * [edge](https://vimcolorschemes.com/sainnhe/edge)
    * [minimalist](https://vimcolorschemes.com/dikiaap/minimalist)
    * [nord-vim](https://vimcolorschemes.com/arcticicestudio/nord-vim)
    * [iceberg.vim](https://vimcolorschemes.com/cocopon/iceberg.vim)

---

Fwiw I've made [few](https://github.com/cespare/vim-toml/pull/56) [PRs](https://github.com/uiiaoo/java-syntax.vim/pull/6) to syntax files following this theme. 

I also wrote [my own colorscheme](https://github.com/Iron-E/nvim-highlite) that works with any color range (16, 256, truecolor) but requires Neovim.

Iron-E:
@da-x re: https://github.com/rust-lang/rust.vim/pull/431#issuecomment-912936599

I made a [post on reddit](https://redd.it/qxrph3) nine days ago to look for testers. With these three new tests, and the two previously, plus you and I, that makes seven.

What are your thoughts?

kevinclark:
@Iron-E did want to mention I moved to soluarized and haven't had any issues! Did a little tweaking, but that’s a feature not a bug.

Iron-E:
That's great to hear! Every once and a while someone mentions that project and makes me glad I ported it.

da-x:
@Iron-E, I just found something an issue with number highlighting on my [keytree](https://github.com/da-x/keytree) project -
image
From https://github.com/da-x/keytree/blob/7e1f43fb60b778f9ad36338cbac14b716ae80f1b/src/main.rs#L80

But apart from that, I think we'll merge this soon enough!

Iron-E:
That issue should be fixed. It thought that the `|` was an indication of a closure, rather than an operation.

> apart from that, I think we'll merge this soon enough!

Glad to hear it!

kevinclark:
@Iron-E I've got an issue where `#[test]` is being detected as rustGenericRegion. I pulled from your branch and it didn't fix it and now my inlayHints (via coc-rustanalyzer) have now changed to white from colored. Code is [here](https://gist.github.com/kevinclark/35868ab481b2bf58747f98f50c1fc486).


Screen Shot 2021-12-03 at 11 51 19 AM

Any thoughts?

Iron-E:
I can reproduce it. Thanks for sharing the code— I tried using just the screenshot but needed the full snippet. I'll see what I can do about it

kevinclark:
@Iron-E Possible a CoC update came in at the same time (I was using PlugUpdate). It's probably coincidental.

Iron-E:
The issue should be fixed. Let me know if it's still happening

kevinclark:
@Iron-E It's fixed, thanks!

Iron-E:
Has anyone found any other issues?

chris-morgan:
Oh yeah, I was going to take a detailed look at this a few weeks ago but then got busy with something else. I’ll try it out for a detailed review and comparison tomorrow (just about to go to bed now). Sorry for letting this drag on so long.

chris-morgan:
Unstructured impressions from a little poking and prodding:

- `rustStaticLifetime` should not `contains=rustStorage`. The `static` keyword means something completely different.
- I strongly prefer `hi def link rustStaticLifetime rustLifetime`; I don’t think it warrants different highlighting, especially not *completely* different highlighting. It’s still a lifetime.
- I’d ditch `rustAnonymousLifetime`; I see no reason why it should be special; but if not, I’d still default it to `rustLifetime`, and not `contains=rustUnused`. Given `rustUnused`, it’s worth pointing out that lifetime `'_` and pattern `_` mean *completely* different things: pattern `_` is unused, but type `_` and lifetime `'_` are inference placeholders. (I don’t think the entire `rustUnused` distinction is worthwhile, actually, and would just leave it with standard pattern or type or identifier or whatever the context is highlighting. I particularly dislike the different highlighting of just the leading underscore in a variable name.)
- `'_a` is a valid lifetime, but `rustAnonymousLifetime` breaks it. If keeping `rustAnonymousLifetime`, it needs a word boundary at its end (`/'_\>/`).
- Why does `@rustLifetimes` contain `rustCharacter`?
- I immediately found a place where you aren’t highlighting comments. This is a bad sign: where there’s one, there are likely to be others. Comments are allowed *everywhere* except inside string literals. This specific case was inside rustGenericRegion. (It’s not terribly common to put comments inside a generic region like this, but it was just your luck that it was done in *[literally the first piece of code that I looked at](https://github.com/launchbadge/sqlx/blob/v0.5.13/sqlx-core/src/postgres/arguments.rs#L35-L36=)!*)
- The way you end up with things like `nextgroup=@rustTokens`… well, this is all a part of why I say that Vim’s syntax highlighting engine is just not suitable for this sort of thing. Break away from the happy path of a-is-followed-by-b-is-followed-by-c by putting a comment in, and bam, the whole thing falls apart. As one example, `||` is *normally* detected as `rustAnonymousFunc` or `rustOperator` correctly, but there’s plenty of perfectly ordinary code that will cause it to fail to detect `rustAnonymousFunc`, where what was there before just never tried to make the distinction. (Sure, we did already do dubious `&&` disambiguation, but there at least normal code would never be misidentified.)
- While on the topic of `rustAnonymousFunc`: rename it `rustClosure`.
- For `rustNoise`, `Delimiter` is not appropriate; it’s documented as “character that needs attention”, and in the general group `Special`; not many syntax files use it for things as pervasive as `rustNoise`—most leave such characters unhighlighted, which is what I would do here. Yet I personally wouldn’t just do it by not linking `rustNoise` to anything, I would instead do it with `:syn-transparent`, because otherwise it loses the highlighting of the parent item, which is important in places like attributes. (Sure, the real problem here is that Vim can’t nest highlighting, e.g. `term=inherit,bold cterm=inherit,bold gui=inherit,bold termfg=inherit ctermfg=inherit guifg=inherit` to add bold without changing italic or colour, but we work with what we’ve got, and that’s a part of the reason why I changed my attribute highlighting from underlined to italic, because it’s less jarring when some characters inside lose the italicness. As it stands, your non-`transparent` `rustNoise` is forcing the parentheses in `#[cfg(not(any(unix, other_prop)))]` to not be italic, while the rest of the line is.)
- `rustStringDelimiter`: if you must separate it out at all (I don’t think it’s worthwhile, *even with* `r####"…"####` strings), leave it linked to `rustString` by default, not `Delimiter`, in conformance with the vast majority of syntax files.
- I would say that `rustCharacter` should match `rustString` in its matchgroup setup. To me this argues against retaining `rustStringDelimiter`.
- I don’t think `rustAssert` should be changed to `Debug`. Assertions are not debugging statements—as a generalisation, I’d say that debugging statements are things that should never go to production, and assertions (even debug assertions) are not that. I’d return it to `PreCondit` (which is just different enough to `Macro` to possibly be useful, while still being of the same group so that it’s similar enough, especially in light of the fact that other crates could have assert macros; hmm… I wonder about generalising `rustAssert` to any macro name that includes “assert”).
- I don’t like `\v` in general because I’m used to reading normal Vim regular expression syntax, and it messes with that, *even if* it’s by bringing it closer to other regex engine syntaxes. *Mixing* `\v` and not-`\v` is even worse.
- There are a whole lot of cases of dubious or bad reuse. `rustBounds`, for example, is a colon and is used where colon is followed by trait bounds, a type name, a macro variable, and nothing as part of a label. `rustMacroVariable` is *indiscriminately* followed by `rustBounds`, triggering potentially inappropriate highlighting of stuff after the colon, rather than structural macro rules using a macro-variable-with-type in the matcher and just-macro-variable in the transcriber. (I know Vim’s syntax engine makes *truly* implementing it in the way I describe hard and sometimes incompatible with other goals here, but at the least I want to point it out.) `rustBounds` should either be called `rustColon`, admitting that it’s just representing the colon token, or be split into at least three (`rustColonBounds`, `rustColonType`, `rustColonMacroFragSpec`, and possibly `rustLabelColon` though I’d just ditch that case).
- `Include` is not suitable for `rustModule`. I doubt anything *is* suitable. (It’d be fair game for the `mod` keyword and `include!`, and perhaps even `include_str!` and others like it, but that’s about it.)
- `rustDerive` doesn’t highlight module paths like in `#[derive(serde::Deserialize)]`. Again, this is the sort of case that concerns me about the entire approach of trying to be clever about what’s allowed where rather than just not bothering. (It’s seldom going to be a *big* deal, but y’know, the principle of the thing. 🙂)
- `rustCfg` should be contained. Also syn-keyword is sufficient and suitable for it, no need for syn-match.
- `cfg!()` is mishighlighted.
- `Option` and `Result` aren’t contained, but other prelude items are.
- Once you’re going in for fancier highlighting, I’m not fond of `type Foo = Bar;` highlighting `Foo` and `Bar` the same.
- Label highlighting is broken.
- In test/sample.rs, you’ve got lots of commas that should be semicolons. (I figure the file should probably parse.)

Iron-E:
I appreciate the feedback! I'll try to address each point with commits and if I have any questions I'll let you know.

Iron-E:
I have addressed most of the concerns which you noted in the previous commits. There are only a few questions/comments I had which I am hoping you could clarify/discuss with me further:

> As one example, `||` is _normally_ detected as `rustAnonymousFunc` or `rustOperator` correctly, but there’s plenty of perfectly ordinary code that will cause it to fail to detect `rustAnonymousFunc`, where what was there before just never tried to make the distinction. (Sure, we did already do dubious `&amp;&amp;` disambiguation, but there at least normal code would never be misidentified.)

The reason why I opted for `nextgroup=@rustTokens` is that I specifically ordered the `syn cluster rustTokens contains=` in so that unless for 99% of cases it would parse correctly (since Vim prefers syntax groups earlier in the `contains`). This is partly because in Rust, as we both know, most tokens can follow most other tokens. In cases where this does not hold, you will see snippets like this:

```vim
syn match rustModule "\v(r#)?<\l(\l|\d)*(_+(\l|\d)*)*>" contained contains=rustRawModule,@rustScopes display
```

As for closures specifically, could you give an example of such breakage? There is a whole section of the sample file dedicated to testing this exact case.

cap

> For `rustNoise`, `Delimiter` is not appropriate; it’s documented as “character that needs attention”, and in the general group `Special`; not many syntax files use it for things as pervasive as `rustNoise`—most leave such characters unhighlighted, which is what I would do here. Yet I personally wouldn’t just do it by not linking `rustNoise` to anything, I would instead do it with `:syn-transparent`, because otherwise it loses the highlighting of the parent item, which is important in places like attributes. (Sure, the real problem here is that Vim can’t nest highlighting, e.g. `term=inherit,bold cterm=inherit,bold gui=inherit,bold termfg=inherit ctermfg=inherit guifg=inherit` to add bold without changing italic or colour, but we work with what we’ve got, and that’s a part of the reason why I changed my attribute highlighting from underlined to italic, because it’s less jarring when some characters inside lose the italicness. As it stands, your non-`transparent` `rustNoise` is forcing the parentheses in `#[cfg(not(any(unix, other_prop)))]` to not be italic, while the rest of the line is.)

This is one area I disagree. The `rustNoise` characters are [delimiters](https://www.merriam-webster.com/dictionary/delimiter), which Merriam-Webster says are "character[s] that mark the beginning or end of a unit of data". Parentheses, commas, and the like _are_ delimiters in this way. They do need attention to show the delimitation of statements. `#[cfg(not(any(unix, other_prop)))]` looks how I expect with regard to parens— the only reason `#[]` doesn't follow the same pattern is that it is a markedly different construct.

> I don’t like `\v` in general because I’m used to reading normal Vim regular expression syntax, and it messes with that, <em>even if</em> it’s by bringing it closer to other regex engine syntaxes. _Mixing_ `\v` and not-`\v` is even worse.

`\v` is closer to Rust's own regex crate, which is why I opted for it. The expressions which do not use `\v` were either not written by me or were shorter without it (since I am going for readability). `:h \v` provides a table which makes reading both `\v` and `\V` side-by-side — despite being opposites — clear.

> In test/sample.rs, you’ve got lots of commas that should be semicolons. (I figure the file should probably parse.)

Could you note an example for this as well?

> `rustMacroVariable` is _indiscriminately_ followed by `rustBounds`, triggering potentially inappropriate highlighting of stuff after the colon

I don't necessarily think the use of `rustBounds` is inappropriate. `:` is conceptually a binding character, which applies broadly from everything to parameters of a function to their types, generics to their bounds, macro variables to a token category, labels to a scope. Can you give an example of such a breakage in this case?

mibmo:
What's stopping this from getting merged? :)

Iron-E:
@chris-morgan or @da-x any updates?

Iron-E:
@chris-morgan @da-x checking in again— what is the status of this pull? Is there anything blocking it?

da-x:
@chris-morgan only waiting your ack, being it a big change around highlighting.
I've been using it for awhile and happy.

ModProg:
I feel like the keywords in macro_rules could be actually highlighted maybe (i.e. pub fn self...).
The current branch:
grafik

Master:
grafik

PR #459, opened 2021-09-24
Set b:undo_indent

dkearns:
See :help undo_indent for details.


chris-morgan:
Thanks for this; I’ve landed the intent of this patch in 4aa69b84c8a58fcec6b6dad6fe244b916b1cf830 (differences: `<lisp` needed to be `lisp<`, and changing the names to the full forms which we use, and order to match).

dkearns:
Sorry about the typo.  I was updating all distributed indent files but it didn't seem worth automating until I'd already done most of them. ;)

There was also no consistent style and mixed seemed quite common so I stopped guessing.

Thanks for the excellent Vim support.

PR #476, opened 2022-05-25
Adds rustStringIdentifier region

sanpii:
The RFC [implicit named arguments for formatting macros](https://github.com/rust-lang/rfcs/pull/2795) is merged since rust 1.59, I think it’s more readable to highlight the identifier.

Before:
2022-05-25-210840_1676x963_scrot
After:
2022-05-25-210819_1676x981_scrot

It’s my first vim syntax contribution, it’s probably perfectible.


chris-morgan:
I’ve thought about this sort of thing before (prototyped it a few years ago), but overall I don’t like it because it’s too faulty: the use of curly braces don’t mean that it’s a format string. If this were applied only to certain whitelisted macros (e.g. format, print, eprintln, panic, todo, assert’s *second* argument which is not terribly easy to express in Vim syntax highlighting, and in practice not even worth trying), there would be cause to do it, but more generally I don’t think it’s a good idea.

On the specific highlighting, I’d want at least { and } to be something that defaulted to Special, and very probably the format specifier from the colon onwards; {{ and }} would also need to be Special, as an escape sequence. And I don’t think Identifier is the right group to use at all. Special is probably the least bad choice for a default highlight link.

(Incidentally, I’d say RFC 2795 is a red herring here.)

PR #477, opened 2022-06-17
Fix `$(...)?` highlighting

rhysd:
`$(...)?` is a macro repeat operator (0 or 1) like `$(...)*` (0 or more) or `$(...)+` (1 or more). `+` and `*` are correctly highlighted, but `?` is not highlighted. This PR adds highlight for `$(...)?`. `*` and `+` allows delimiter like `$(...),*`, but `?` doesn't.

Before:

before

After:

after


Iron-E:
Already fixed by #431

PR #485, opened 2022-09-08
Better errorformat

gen740:
Rust compiler output
```
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
did not ignored due to errorformat was set `\%Inote:\ %m` first.
So, change `+=` to `^=` solve this problem.

And, more `cargo test` cmmand output `error: test failed, to rerun pass ...` should be ignored.

I wish this change would apply to vim repository.
In vim repository,  `compiler/cargo.vim` is not up-to-date, so we cannot parse error correctly.

### DIFF
--- a/compiler/cargo.vim
+++ b/compiler/cargo.vim
@@ -32,14 +32,14 @@ augroup RustCargoQuickFixHooks
 augroup END
 
 " Ignore general cargo progress messages
-CompilerSet errorformat+=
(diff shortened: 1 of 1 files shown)


gen740:
I found similar pull request #445.
So, I decided to take in these changes to this pull request.

gen740:
Close this PR because master's `cargo.vim` was updated.

Issue #486, opened 2022-10-23
Why is `as` highlighted like an operator and not a keyword?


mweber15:
This has been done this way for a very long time (2111663c), but it's unclear (to me) why. The result is that Vim highlighting doesn't match highlighting within the book, and possibly other places as well, which is a bit jarring. If I change this then I get the result that I expect in the specific scenario where I noticed it, but I assume there's a good reason for the current behavior.

* rust.vim version:
```
% pwd
/home/matt/.vim/pack/plugins/start/rust.vim
% git log -1 --oneline
1cdc5cb (HEAD -> master, origin/master, origin/HEAD) Ignore `For more information` line
```

Steps to reproduce:
Use the `as` keyword anywhere.

Expected vs. actual behavior:

Expected: keyword highlighting
![image](https://user-images.githubusercontent.com/30441572/197401822-b6cce7a4-a0ff-420f-b0af-657da60709fa.png)

Actual: operator highlighting
![image](https://user-images.githubusercontent.com/30441572/197401863-3edadaf6-96d0-45e6-89d5-f7a6d0560241.png)

Paste debugging info from the Rust Vim plugin via _one_ of the following
commands: `:RustInfo`,  `:RustInfoToClipboard`, or `:RustInfoToFile <filename>`.

```
rust.vim Global Variables:

let g:ftplugin_rust_source_path = v:null
let g:loaded_syntastic_rust_cargo_checker = v:null
let g:loaded_syntastic_rust_filetype = v:null
let g:loaded_syntastic_rust_rustc_checker = v:null
let g:rust_bang_comment_leader = v:null
let g:rust_cargo_avoid_whole_workspace = v:null
let g:rust_clip_command = v:null
let g:rust_conceal = v:null
let g:rust_conceal_mod_path = v:null
let g:rust_conceal_pub = v:null
let g:rust_fold = v:null
let g:rust_last_args = v:null
let b:rust_last_args = []
let g:rust_last_rustc_args = v:null
let b:rust_last_rustc_args = []
let g:rust_original_delimitMate_excluded_regions = v:null
let g:rust_playpen_url = v:null
let g:rust_prev_delimitMate_quotes = v:null
let g:rust_recent_nearest_cargo_tol = v:null
let g:rust_recent_root_cargo_toml = v:null
let g:rust_recommended_style = v:null
let g:rust_set_conceallevel = v:null
let g:rust_set_conceallevel=1 = v:null
let g:rust_set_foldmethod = v:null
[...]

PR #487, opened 2022-11-06
Allow option to highlight operator `as` as keyword

mweber15:
While `as` is functionally an operator, it is often highlighted as if it were a keyword. This change gives the user the option to link `rustOperatorKeyword` to `Keyword` while preserving the current default behavior. This allows for `as` to be highlighted like a keyword separately from other operators.

The documentation lists `as` as a keyword and consistently highlights it as such.

https://doc.rust-lang.org/std/index.html#keywords
https://doc.rust-lang.org/std/keyword.as.html


mweber15:
This is following up on #486.

Iron-E:
This is implemented in #431

Issue #468, opened 2022-01-28
[Feature Request] Highlight integer literal suffixes differently


alexmozaidze:
It would be cool to have integer literal type suffixes highlighted differently. Why would this be useful? Consider this `15312f32`, the `f32` at the end gets jumbled with the numbers, making it harder to read. A workaround to this problem would be to use the holy underscore, but then it makes the expression longer by *one character*, truly disastrous! Please add a special highlighting.

yorik1984:
https://github.com/rust-lang/rust.vim/pull/490

PR #490, opened 2022-12-11
Highlight integer literal suffixes differently #468

yorik1984:


yorik1984:
tbLsES6G9a
marked different colors only for visual effetcts

PR #498, opened 2023-05-07
Support `$()?` in macro_rules

ModProg:


theKidOfArcrania:
would be nice to get this merged... code folding breaks in potentially catastrophic ways if you have $(...)? in one of your `macro_rules!`

PR #511, opened 2024-05-12
Add `RustTest` support for `tokio::test`

bbannier:
This patch adds support for discovering tokio tests with a `tokio::test` macro.


bbannier:
Looking through the list of open PRs, this might just be a dup of #420 or #423.

bbannier:
Closing since this is a dup.

